Issue #832, opened 2022-01-17
Some fonts broken on website


blushi:
**Describe the bug**
Some text doesn't have the right font (Lato) on the website after https://github.com/regen-network/regen-web/pull/797

**To Reproduce**
1. Go to https://www.regen.network/
2. Look at the fonts in the top menu dropdowns or the footer, and some other parts like on the homepage:
![image](https://user-images.githubusercontent.com/3149358/149788839-340afeb3-6211-4e59-af5c-8a8eb8908edc.png)

**Expected behavior**
The font should be **Lato**.

**Screenshots**
![image](https://user-images.githubusercontent.com/3149358/149789081-1036ea78-628f-48d5-b004-2fe48a61550c.png)

![image](https://user-images.githubusercontent.com/3149358/149789229-82fc2ce3-cce1-4d07-9804-70f26100f369.png)

cc/ @haveanicedavid 

____

## For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels and zenhub epics applied
- [x] Appropriate contributors tagged

PR #835, opened 2022-01-18
Set root font to lato

haveanicedavid:
## Description

Closes: #832 

Before the sanity migration (#797 ) we handled base styles differently in /registry and /www.  

I normalized them into the `RegenThemeProvider` both apps use in order to troubleshoot another styles-related bug (MUI styles being overridden on build) and in doing so deleted web-www's [layout.css](https://github.com/regen-network/regen-web/blob/cfa4b1506c3b293470c9a0d89633ba3c69ce4f79/web-www/src/layouts/layout.css) file. I assumed this was serving the same purpose as MUI's `CSSBaseline` (which registry is using but the Gatsby site was not) and thought by moving CSSBaseline to the theme, I wouldn't need `layout.css`, but didn't notice the registry `index.css` set the default to apple system font, but the website layout was set to 'Lato'.

Everything _should_ be the same already with the two apps because our `muiTheme.ts` file sets the base font to lato (and it works on the registry site), but I suspect the same build issue I mentioned above might be messing with style hierarchy. 

If we're concerned about other ramifications I can restore the deleted `layout.css` file, but I don't think it's ideal to maintain separately long-term if we want to keep style parity across the two apps and I believe this PR fixes the immediate problem. I could also move the prior contents of `layout.css` into the shared theme styles, but that seems more likely to cause side-effects and I figure we don't want to maintain the extra CSS if we don't have to

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

- [ ] reviewed tests


blushi (approved):
Thanks for fixing this

mhagel (approved):

PR #838, opened 2022-01-19
Fix some minor styles

victordelval:
## Description

Closes: regen-network/regen-registry#797

Some styling details on the Credit Batch table and section:

- font size to 16px in table body cells, 
- table row height to 100px, 
- change background image (lighter contour lines)

All sections in page:

- titles of the sections aligned to the center

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

- [ ] reviewed tests


web-registry/src/components/organisms/CreditBatches.tsx
```diff
@@ -40,13 +40,13 @@ interface HeadCell {
 }
 
 const headCells: HeadCell[] = [
+  { id: 'start_date', numeric: true, label: 'date created' },
```
mhagel, line 43: This is in the comps, but we removed it because `start_date` != 'date created'. "Date created" is not yet available in the blockchain data. 
```suggestion
```

web-registry/src/components/organisms/CreditBatches.tsx
```diff
@@ -40,13 +40,13 @@ interface HeadCell {
...
   { id: 'issuer', numeric: false, label: 'issuer' },
   { id: 'batch_denom', numeric: false, label: 'batch denom' },
   { id: 'class_id', numeric: false, label: 'credit class' },
-  { id: 'tradable_supply', numeric: true, label: 'total amount tradable', wrap: true },
+  { id: 'tradable_supply', numeric: true, label: 'total amount active', wrap: true },
```
mhagel, line 47:
```suggestion
  { id: 'tradable_supply', numeric: true, label: 'total amount tradable', wrap: true },
```
This was intentional, as "active" is not really accurate here. I know the mocks are different.

web-registry/src/components/organisms/CreditBatches.tsx
```diff
@@ -40,13 +40,13 @@ interface HeadCell {
...
   { id: 'retired_supply', numeric: true, label: 'total amount retired', wrap: true },
   { id: 'amount_cancelled', numeric: true, label: 'total amount cancelled', wrap: true },
-  { id: 'start_date', numeric: true, label: 'start date' },
```
mhagel, line 49: bring this back here please

web-registry/src/components/organisms/CreditBatches.tsx
```diff
@@ -164,6 +159,10 @@ const CreditBatches: React.FC = () => {
             {stableSort(batches, getComparator(order, orderBy)).map((batch: any) => {
               return (
                 <StyledTableRow className={styles.noWrap} tabIndex={-1} key={batch.batch_denom}>
+                  <StyledTableCell>
```
mhagel, line 162: See comment above
```suggestion
```

web-registry/src/components/organisms/CreditBatches.tsx
```diff
@@ -174,7 +173,6 @@ const CreditBatches: React.FC = () => {
                   <StyledTableCell>{formatNumber(batch.tradable_supply)}</StyledTableCell>
                   <StyledTableCell>{formatNumber(batch.retired_supply)}</StyledTableCell>
                   <StyledTableCell>{formatNumber(batch.amount_cancelled)}</StyledTableCell>
-                  <StyledTableCell>{moment(batch.start_date).format('LL')}</StyledTableCell>
```
mhagel, line 177: keep

web-registry/src/pages/Home.tsx
```diff
@@ -10,14 +10,14 @@ import { ProjectCards, CreditClassCards, CreditBatches } from '../components/org
 import { creditClasses } from '../mocks';
 
 import cowsImg from '../assets/cows-by-barn.png';
-import topographyImg from '../assets/background.jpg';
+import topographyImg from '../assets/background-contour-1.jpg';
```
mhagel, line 13: Any specific reason for adding this file? Better quality?

mhagel (changes requested):
Good catches, but most of these were different from the figma mockups for reasons that came up right before you joined. Maybe @erikalogie can update the mocks for us?

Apologies for not communicating this better, but thanks for your attention to detail!

web-registry/src/pages/Home.tsx
victordelval: It is the one that corresponds to the design in Figma currently. The lines are a little clearer, so I guess the visual result is a little better, since the table is also in light gray tones, so the background stands out a little less.

web-registry/src/components/organisms/CreditBatches.tsx
victordelval: thanks!

web-registry/src/pages/Home.tsx
mhagel: Ok. We probably only want to have one image file for this purpose. We may use `background.jpg` elsewhere in the app. If `background-contour-1.jpg` is better, we should use it in those cases too.

However, I think the difference may be minimal, and adding this file is probably unnecessary for this issue. To avoid scope creep, I recommend you just keep the original.

Alternatively, you could send Erika the deploy-preview link and ask her opinion.

mhagel (approved):

web-registry/src/pages/Home.tsx
```diff
@@ -86,7 +86,7 @@ const Home: React.FC = () => {
       />
 
       {projectsData?.allProjects?.nodes && (
-        <Section title="Projects" titleAlign="left" classes={{ root: styles.section, title: styles.title }}>
+        <Section title="Featured Projects" classes={{ root: styles.section, title: styles.title }}>
```
blushi, line 89: we display all projects by default right now (just fetching all existing projects in the pg database) so if we want to only display some featured projects in the future, we'll need an issue to do that, it could be configurable through Sanity for instance cc/ @erikalogie (although maybe that's not so urgent given we don't have a ton of projects yet)

erikalogie:
Yeah I don't think configuring it is urgent quite yet because we have so
few projects, but I agree Sanity would be great for that.

victordelval:
The build process has failed several times. It was a symbolic PR, more of a test than anything else, with 2 minor style improvements to be included elsewhere. It's not worth spending another minute..

blushi:
@victordelval Looking at netlify logs, it seems related to the fact that the cache had to be cleared for web-www and web-registry to get built successfully (that's because the initial deploy previews were using the pre-upgrade version and now it's using the post-upgrade version).
I would re-open this and re-run those with cleared cache.

PR #855, opened 2022-02-28
View my Credit Basket tokens [UI-only]

blushi:
## Description

Closes: https://github.com/regen-network/regen-registry/issues/813

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

- [ ] reviewed tests


web-components/src/components/table/ActionsTable.tsx
```diff
@@ -0,0 +1,183 @@
...
+    left: 0,
+    bottom: 0,
+    borderLeft: `1px solid ${theme.palette.info.light}`,
+  },
+  greyText: {
+    color: theme.palette.info.main,
+  },
+}));
+
+export type renderActionButtonsFunc = (i: number) => React.ReactElement;
```
haveanicedavid, line 39:
```suggestion
export type RenderActionButtonsFunc = (i: number) => React.ReactElement;
```
I'm a bit confused about what the index is for - just to pass an index down to be used as `key`?

web-registry/src/components/organisms/BasketsTable.tsx
```diff
@@ -0,0 +1,61 @@
...
+              </Box>
+              <Box sx={{ color: 'info.main' }}>
+                {row.display_denom.toLocaleString()}
+              </Box>
+            </Grid>
+          </Grid>,
+          row.balance?.amount
+            ? parseInt(row.balance?.amount) /
+              Math.pow(10, parseInt(row.exponent))
+            : 0,
```
haveanicedavid, line 56: are these numbers the same? if so and `exponent` is always available, maybe we should just use that?

haveanicedavid (approved):
LGTM!

web-components/src/components/table/ActionsTable.tsx
blushi: to get the corresponding row and inform the "action modal" that will open up (eg if I want to retire some ecocredits, we should provide the batch denom and max amount that the user can retire to the modal, and we can get this using `row[i]` info)

web-registry/src/components/organisms/BasketsTable.tsx
blushi: a basket token always has an exponent defined, let's called it `exp`, meaning that for `x` ecocredits you put into this basket, you then get `x * 10^exp` basket tokens back. 
these basket tokens are tracked as regular coins (just like `regen`s) in the bank module as integers so when you query someone's balance for this basket token, you'll get the amount `x * 10^exp`. For the rNCT token, the corresponding denom will be `eco.uC.rNCT` (notice the `u` which corresponds to `micro`, meaning the exponent is 6, just like in `uregen`). But on the UI we actually want to display the amount in `eco.C.rNCT` which corresponds to the initial amount of ecocredits so we need to divide the amount we get in `eco.uC.rNCT` from the bank module by `10^exp`

See the docs on regen-ledger: https://github.com/regen-network/regen-ledger/blob/release/v3.0.x/proto/regen/ecocredit/basket/v1/tx.proto#L44-L68

victordelval (approved):

web-registry/src/components/organisms/BasketsTable.tsx
haveanicedavid: That makes sense - thanks for the explanation!

PR #856, opened 2022-03-03
Victor/814 basket details page

victordelval:
## Description

Closes: regen-network/regen-registry#814

Adds the first version of the basket details page (`BasketDetails`)

Mocked and hardcoded data are used.

It presents an upper overview section (`BasketOverview`) and the ecocredits section (`BasketEcocredits`) with the table that shows the credits batches included in the basket. 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

- [ ] reviewed tests


blushi:
@victordelval is this ready for review now?

web-registry/src/components/organisms/BasketEcocredits.tsx
```diff
@@ -0,0 +1,60 @@
...
+  }, [basketDenom]);
+
+  return (
+    <Section
+      title="Ecocredits"
+      titleVariant="h3"
+      titleAlign="left"
+      classes={styles}
+    >
+      <EcocreditsTable credits={credits} />
```
blushi, line 55: you can't use the ecocredits table component as is, since it has "Amount Tradable" and "Amount Retired" columns while there's only "Amount" on the table on the basket details page (since amount retired doesn't make any sense in this context)

web-registry/src/components/organisms/BasketOverview.tsx
```diff
@@ -0,0 +1,274 @@
...
+export { BasketOverview };
+
+/**
+ * Basket summary item (subcomponent)
+ */
+
+const useStylesItem = makeStyles(theme => ({
+  gridItem: {
+    marginTop: theme.spacing(4),
+  },
```
haveanicedavid, line 225: this is maybe better for an engineering call (or github discussion like we discussed previously), but I think we might want to avoid using `makeStyles` as much as possible moving forward, as it won't work in React 18 (see the disclaimer at the top of [this page](https://mui.com/styles/basics/)

For a bit of context, we were on MUI v3 until recently (skipped 4, went straight to 5) and with MUI v3, `makeStyles` was the only styling option. It uses JSS under the hood, but MUI switched to Emotion with v5 and JSS is now deprecated. 

I've been playing with the options some, and my personal preference is using the [sx prop](https://mui.com/system/the-sx-prop/#main-content) for positioning (padding, margin etc) and MUI's [styled](https://mui.com/customization/how-to-customize/#2-reusable-style-overrides) for reusable styles. 

There will also be places where it won't make sense to use the new options (mainly when interacting with older components which accept `classes` or a JSS `className` as a prop you need to pass - as you've done several places in this story), and It might be annoying to have 3 forms of styles in the same codebase (`styled`, `makeStyles`, and `cx`), but they're all connected and assuming MUI doesn't change their plans moving forward, we'll eventually have to migrate away from `makeStyles` to upgrade react, and using the newer syntax now might help with that.

Again, this is probably not the best place to discuss, but I wanted to mention given your recent push for code quality maintenance.

web-registry/src/components/organisms/BasketOverview.tsx
```diff
@@ -0,0 +1,274 @@
...
+  label: string;
+  data: string | JSX.Element;
+  link?: string;
+}
+
+const Item = ({ label, data, link }: ItemProps): JSX.Element => {
+  const styles = useStylesItem();
+
+  return (
+    <div className={styles.gridItem}>
```
haveanicedavid, line 260: To put my longer comment in context with an example, this element could be rewritten as
```suggestion
    <Box cx={{ mt: 4 }}>
```

and you could remove the `gridItem` styles. 

alternatively, you could make a 
```tsx
const GridItem = styled('div')(({ theme }) => ({
  marginTop: theme.spacing(4),
}))
```

and change this line to 
```suggestion
    <GridItem>
```

but that's the same amount of boilerplate, so I prefer `sx` for styles that aren't reused (it's a good bit less code)

To be clear with these comments: I'm not saying you should remove all of the `makeStyles`s in this story (you would probably still need some to pass `classes` to components like the `<Section>` element in `BasketOverview`), and am fine with folks using `makeStyles` if that's the preference, but do think it could create some dependency friction in the future when we go to upgrade

web-registry/src/components/organisms/BasketEcocredits.tsx
haveanicedavid: ^ You could pull this out to a `BasketEcocreditsTable` component which consumes `<ActionsTable>` in the same way the `<EcocreditsTable>` does currently (but with the columns changed)

haveanicedavid (commented):
Looking really good so far! I've used this story to bring up some opinions on styles moving forward, but think this will be good to go once the Ecocredits table is swapped out. Great work!

web-registry/src/components/organisms/BasketOverview.tsx
victordelval: I take note, thanks a lot!

blushi:
@victordelval any updates here?

victordelval:
> 

The part corresponding to data management, interfaces and the new table component has been updated. I need to improve the styles according to the comments.

web-registry/src/components/organisms/BasketOverview.tsx
victordelval: Wow, I've just been testing a bit and it's really tricky. Now I am going to make the minimum changes, since the adaptation of existing components is not so strict, and we will make this topic a priority for the coming weeks. Let's analyze the different cases and agree among all the best ways to apply the styles and the changes now to prepare for the following refactorings to come. Thanks, David!!

victordelval:
By the way @haveanicedavid I have extracted 2 mini components, atoms :-) from the ecocredits table, which I also used here. GreyText and BreakText (the latter I have renamed, instead of Amount)

web-registry/src/components/organisms/BasketOverview.tsx
```diff
@@ -0,0 +1,268 @@
...
+      <Section className={styles.content}>
+        <Grid container>
+          <Grid item xs={12} sm={5} className={styles.imageContainer}>
+            <OptimizedImage
+              className={styles.image}
+              src={forestImg}
+              alt={name}
+            />
+          </Grid>
+          <Grid item xs={12} sm={7} className={styles.textContainer}>
```
haveanicedavid, line 138: looks like many of these `makeStyles` classes aren't necessary. 

Could either handle with `sx` (which might take some reworking because of how `sx` handles breakpoints - they're all 'up' e.g. `<Grid sx={{ p: { xs: 10, sm: 12 } }}>` translates to `[theme.breakpoints.up('xs')]: padding: theme.spacing(10)` etc), or by making these with `styled`:
```tsx
const TextContainer = styled(Grid)(({ theme }) => ({
  display: 'flex',
  flex: 1,
  flexDirection: 'column',
  [theme.breakpoints.down('lg')]: {
    paddingLeft: theme.spacing(8),
  },
  [theme.breakpoints.down('sm')]: {
    width: '100%',
    padding: theme.spacing(8, 0),
    '&:last-child': {
      paddingBottom: theme.spacing(15),
    },
  },
}));
```

then you'd instantiate it as
```suggestion
          <TextContainer item xs={12} sm={7}>
```
etc

web-registry/src/components/organisms/BasketOverview.tsx
```diff
@@ -0,0 +1,268 @@
...
+}) => {
+  const styles = useStyles();
+
+  return (
+    <Box
+      sx={theme => ({
+        display: 'flex',
+        justifyContent: 'center',
+        borderBottom: `1px solid ${theme.palette.grey[100]}`,
+        backgroundColor: theme.palette.grey[50],
```
haveanicedavid, line 123: You might already know this, but the `theme` callback isn't necessary here, you can also write this as 
```suggestion
        backgroundColor: 'grey.50'
```
though it'd take splitting `borderBottom` into two properties...
```tsx
borderBottom: 1,
borderColor: 'grey.100'
```

web-registry/src/components/organisms/BasketOverview.tsx
```diff
@@ -0,0 +1,268 @@
...
+              </Grid>
+            </OnBoardingCard>
+          </Grid>
+        </Grid>
+      </Section>
+    </Box>
+  );
+};
+
+export { BasketOverview };
```
haveanicedavid, line 198: let's either export the function directly ie
```tsx
export const BasketOverview: React.FC<BasketOverviewProps> = ({
```

or move all named exports to the bottom of the file, for consistency

web-registry/src/components/organisms/BasketOverview.tsx
```diff
@@ -0,0 +1,268 @@
...
+              {parseText(data)}
+              <SmallArrowIcon className={styles.arrowIcon} />
+            </Link>
+          ) : (
+            parseText(data)
+          )}
+        </Description>
+      )}
+    </Box>
+  );
```
haveanicedavid, line 267: if you force `data` into an array, you can DRY up the JSX:
```tsx
const itemData = Array.isArray(data) ? data : [data];
```
then you could 
```suggestion
      {itemData.map((item, i) => (
        <Description key={`basket-${item}`} className={styles.data}>
          {link ? (
            <Link href={link + item} target="_blank" rel="noreferrer">
              {parseText(item)}
              <SmallArrowIcon className={styles.arrowIcon} />
            </Link>
          ) : (
            parseText(item)
          )}
        </Description>
      ))}
```

web-registry/src/components/organisms/EcocreditsTable.tsx
```diff
@@ -10,18 +10,11 @@ import { formatDate } from 'web-components/lib/utils/format';
 
 import { truncate } from '../../lib/wallet';
 import { getAccountUrl } from '../../lib/block-explorer';
+import { GreyText } from '../atoms/GreyText';
+import { BreakText } from '../atoms/BreakText';
```
haveanicedavid, line 14: while I love the idea of moving these to be reusable, it might be preferable to keep simple stuff like this within the context of the component it's used. 

For example, I might see these and want to combine the two, which could be done:
```tsx
<GreyText><BreakText>this text will be grey and break on word</BreakText></GreyText>
```
but the JSX would bloat quickly. We also have several forms of `grey` (grey.50, grey.100, info.main etc), and it's not clear which one this uses. 

However if there were something that was commonly used which required several styles, like... if we had some custom text that all tables use, it could make sense to have a `TableText` atom. 

Does that make sense? I love the idea, but worry it might get a bit confusing over time

haveanicedavid (commented):
A few comments and suggestions, but it's looking good so far!

web-registry/src/types/ledger/ecocredit.ts
```diff
@@ -72,3 +72,70 @@ export interface QueryBasketsResponse {
   baskets: Basket[];
   pagination?: PageResponse;
 }
+
+export interface IBasket {
```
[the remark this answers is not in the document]
victordelval: this is just a temporary thing until we update the regen-js api package. I leave it as it is, it will be removed next week.

without intention to introduce new nomenclature. We'll talk about that in due course, about naming conventions, etc.

initially I had that interface defined with the name Basket, with camelCase, but then another one came in with snake_case, and I simply renamed mine

web-registry/src/types/ledger/ecocredit.ts
```diff
@@ -72,3 +72,70 @@ export interface QueryBasketsResponse {
...
+  balance: string;
+  batchStartDate?: string | Date;
+}
+
+/**
+ * Ecocredits
+ */
+
+/** BatchInfo represents the high-level on-chain information for a credit batch. */
+export interface IBatchInfo {
```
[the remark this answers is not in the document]
victordelval: precisely, after updating master I had to rename the interfaces that I was using. But the components that depend on the interface according to the api are already in camelCase.

I'll leave it as it is, it's temporary

web-registry/src/types/ledger/ecocredit.ts
```diff
@@ -72,3 +72,70 @@ export interface QueryBasketsResponse {
...
+  // pagination?: PageResponse; // DISABLED
+}
+
+// BasketBalance stores the amount of credits from a batch in a basket
+export interface BasketBalance {
+  basketId: string | Long;
+  batchDenom: string;
+  balance: string;
+  batchStartDate?: string | Date;
+}
```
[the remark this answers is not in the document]
victordelval: yes, this is a type coming with the update. let's leave it as it is for the moment, with the update we'll be cleaning it up in the next few days

web-registry/src/hooks/useBasketDetails.ts
```diff
@@ -0,0 +1,110 @@
...
+      }
+    }
+
+    fetchData(basketDenom);
+  }, [basketDenom]);
+
+  // Prepare data for BasketOverview
+  useEffect(() => {
+    if (data?.basket) {
+      const _dataOverview: BasketOverviewProps = {
```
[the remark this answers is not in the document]
victordelval: I think it comes from when I used Python... I use it when the variable represents the "preparation" of a data, something in the process, yes within a delimited scope, or something like that. In this case the final data is called dataOverview (managed by the useState hook). I would have used the same name in the effect hook... so just prefix it with an underscore! I am also taking note of these little details so that over time we can precisely define our own style guide, etc.

web-registry/src/components/organisms/BasketOverview.tsx
victordelval: ok, i just copied the shape from a page component which seemed neat to me. I personally think I would export as default in this case. I take note to define a convention

web-registry/src/components/organisms/BasketOverview.tsx
victordelval: super nice! thanks David!

web-registry/src/components/organisms/EcocreditsTable.tsx
victordelval: I personally see it as temporary. I think we have to do a good definition and refactoring of the Design System and the UI Primitives...

So I don't care, I don't have a preference. I remove the components and return it to its origin then?

web-registry/src/components/organisms/BasketOverview.tsx
```diff
@@ -0,0 +1,268 @@
...
+            )}
+          </Description>
+        ))
+      ) : (
+        <Description className={styles.data}>
+          {link ? (
+            <Link href={link} target="_blank" rel="noreferrer">
+              {parseText(data)}
+              <SmallArrowIcon className={styles.arrowIcon} />
+            </Link>
```
mhagel, line 260: This is a candidate for a new component. It showed up in regen-network/regen-registry#805 as well.

mhagel (approved):
Looks good!

web-registry/src/components/organisms/BasketEcocreditsTable.tsx
```diff
@@ -0,0 +1,93 @@
...
+        // TODO: Project
+        <Box
+          sx={{
+            minWidth: {
+              xs: 'auto',
+              sm: '11rem',
+              lg: '13rem',
+            },
+          }}
+        >
```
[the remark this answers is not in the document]
victordelval: thanks!

web-registry/src/components/organisms/BasketOverview.tsx
victordelval: I hadn't paid enough attention to your comment before, now I realize what you meant. Thanks!

web-registry/src/components/organisms/BasketOverview.tsx
```diff
@@ -0,0 +1,268 @@
...
+import Description from 'web-components/lib/components/description';
+import SmallArrowIcon from 'web-components/lib/components/icons/SmallArrowIcon';
+import { Label } from 'web-components/lib/components/label';
+import { parseText } from 'web-components/lib/utils/textParser';
+import { formatNumber } from 'web-components/lib/components/table';
+import { formatDate } from 'web-components/lib/utils/format';
+
+import { OptimizedImage } from '../atoms/OptimizedImage';
+import topoImg from '../../assets/background-contour-2.svg';
+import forestImg from '../../assets/forest-token.png';
```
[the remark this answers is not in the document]
victordelval: I take note, thanks Mark!

web-registry/src/components/organisms/BasketOverview.tsx
victordelval: done!

blushi:
Just reading the comments re: regen-js upgrade, in particular https://github.com/regen-network/regen-web/pull/856#discussion_r824059652 and there's something I'd like to repeat again because it seems like there is some confusion around that.

The upgrade will indeed make those types available directly from @regen-network/api but we won't actually be using them until we migrate our current queries to use regen-js, which **isn't** something that should be addressed as part of https://github.com/regen-network/regen-registry/issues/850.
I've created a separate issue to track that: https://github.com/regen-network/regen-registry/issues/852 that could eventually be done after we deliver the main basket features as a code-hygiene task.

web-registry/src/types/ledger/ecocredit.ts
```diff
@@ -73,6 +73,73 @@ export interface QueryBasketsResponse {
...
+
+/**
+ * Ecocredits
+ */
+
+/** BatchInfo represents the high-level on-chain information for a credit batch. */
+export interface IBatchInfo {
```
blushi, line 121: I'd rather get rid of the `I` too
plus the interface names are not even consistent anymore, some have `I` while some other don't, which is confusing

blushi (changes requested):

web-registry/src/types/ledger/ecocredit.ts
```diff
@@ -73,6 +73,73 @@ export interface QueryBasketsResponse {
...
+}
+
+export interface QueryBasketResponse {
+  basket?: IBasket;
+  classes: string[];
+}
+
+export interface QueryBasketBalancesRequest {
+  basketDenom: string;
+  // pagination?: PageRequest; // DISABLED
```
blushi, line 100: why disabling them? 
it was already optional

web-registry/src/types/ledger/ecocredit.ts
```diff
@@ -73,6 +73,73 @@ export interface QueryBasketsResponse {
...
+}
+
+export interface QueryBasketBalancesResponse {
+  balances: BasketBalance[];
+  // pagination?: PageResponse; // DISABLED
```
blushi, line 105: ditto

blushi (changes requested):

web-registry/src/components/organisms/BasketEcocreditsTable.tsx
```diff
@@ -0,0 +1,76 @@
...
+  if (!batches?.length) {
+    return <NoCredits title="No credit batches to display" />;
+  }
+
+  return (
+    <ActionsTable
+      tableLabel="basket ecocredits table"
+      renderActionButtons={renderActionButtons}
+      headerRows={[
+        // TODO: Project
```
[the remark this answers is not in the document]
victordelval: That comment is a reminder that that basket credit batch table starts with the column "Project" (project name) which is not implemented (mocked). I guess I should implement mocking the data in the hook and pass it to the component, ok? The fact is that I have seen that the other tables that have been implemented these days do not include that first column either and in the end it has passed me by and I have left it that way. Do I simply remove the comment or add the data by mocking?

web-registry/src/types/ledger/ecocredit.ts
victordelval: ok, as I was commenting, it was a forced renaming by the interfaces that have entered with the new changes, when updating my branch. The interfaces I'm using already have properties in camelCase, but the ones that entered master first go in snake_case. So they are not directly reusable. I just did a trivial rename because I can't reuse those interfaces that have come in without modifying my code or without modifying the existing one. And if we define 2 interfaces with the same name the interface is extended. What should I do?

web-registry/src/components/organisms/BasketEcocreditsTable.tsx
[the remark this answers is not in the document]
blushi: yes I think we should mock the data for now
once we work on pulling the actual data, that will be possible to get the corresponding project for a credit batch through the offchain db

web-registry/src/types/ledger/ecocredit.ts
victordelval: Precisely, since it is optional and we do not use it, and these interface implementations are provisional, I thought that it did not make much sense to import or implement the PageRequest interface (which is currently not there, the one we have available is PageResponse). What should I do?

web-registry/src/types/ledger/ecocredit.ts
blushi: oh ok I get it now, this is because the responses from the REST endpoints come with snake_case while the response from regen-js will comme with camelCase
I think that we should just keep using the REST endpoint for now for querying batch info, so let's just use the existing snake_case `BatchInfo` interface everywhere (and we'll change that later on as part of https://github.com/regen-network/regen-registry/issues/852
Then when it comes to basket, since we haven't implemented any actual queries yet for those, let's just use the camelCase based interfaces everywhere since we'll start using regen-js for querying those.
wdyt?

web-registry/src/types/ledger/ecocredit.ts
blushi: ok fine for now now then

web-registry/src/types/ledger/ecocredit.ts
victordelval: Now I just fully understood it too.. I leave the BatchInfo REST version and update Basket regen-js version. Wonderful Marie, thanks!

web-registry/src/types/ledger/ecocredit.ts
victordelval: As I am finally going to push some changes, it costs me nothing to implement that PageResponse and thus we leave the interfaces of those 2 queries that optionally incorporate it complete. okay?

blushi: sounds good

web-registry/src/types/ledger/ecocredit.ts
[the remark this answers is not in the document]
victordelval: by the way, they finally changed :)

web-registry/src/components/organisms/EcocreditsTable.tsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
-import { Box, styled } from '@mui/material';
+import { Box } from '@mui/material';
+import { styled } from '@mui/material';
```
blushi, line 3: why this change?

web-registry/src/lib/ecocredit.ts
```diff
@@ -213,6 +215,66 @@ export const getTxsByEvent = (msgType: string): Promise<AxiosResponse> => {
...
+      {
+        basketId: 'basket-1',
+        batchDenom: 'C01-20170101-20201231-003',
+        balance: '30000',
+      },
+    ],
+  });
+}
+
+export async function getBatchInfo(
```
blushi, line 261: you could just use the existing `queryEcoBatchInfo` instead of mocking the data

blushi (commented):
looking good but still some small comments

web-registry/src/components/organisms/EcocreditsTable.tsx
victordelval: I was undoing a change. I forgot to check the automatic import. Thanks!

web-registry/src/lib/ecocredit.ts
victordelval: trying to remove the BatchInfo mock I've seen myself changing too many things. I'd rather leave it as it is and move on

web-registry/src/hooks/useBasketDetails.ts
```diff
@@ -0,0 +1,131 @@
...
+      }
+    }
+
+    fetchData(basketDenom);
+  }, [basketDenom]);
+
+  // Prepare data for BasketOverview
+  useEffect(() => {
+    if (data?.basket && data?.basketExtended && data?.batchesInfo) {
+      console.log('>>> _dataOverview', data); // eslint-disable-line no-console
```
blushi, line 99:
```suggestion
```

victordelval: ups

web-registry/src/lib/ecocredit.ts
blushi: Please have a look at https://github.com/regen-network/regen-web/pull/867

victordelval: super cool! thanks Marie!

blushi: you can just merge #867 (the target branch being your branch) if that makes sense to you!

blushi (approved):
LGTM

PR #857, opened 2022-03-04
Simple credit class page

mhagel:
## Description

Closes: regen-network/regen-registry#805
Needs server update https://github.com/regen-network/registry-server/pull/129

If a `classId` is provided in url, like `/credit-classes/C02`, show the simplified credit class page with ledger data, and metadata fetched from postgres db.

The main new code is `CreditClassDetailsSimple`.
Modified a few shared components to make them easier to work with `sx`, among other tweaks.
Added a `metadata-graph` REST route file.
 
---

### Author Checklist

I have...

### Reviewers Checklist

I have...

- [x] confirmed all author checklist items have been addressed
- [x] reviewed code correctness and readability
- [ ] reviewed tests

### DIFF
--- a/web-registry/src/components/atoms/Link.tsx
+++ b/web-registry/src/components/atoms/Link.tsx
@@ -1,23 +1,33 @@
 import React from 'react';
 import { Link as RouterLink } from 'react-router-dom';
 import MuiLink from '@mui/material/Link';
+import { SxProps } from '@mui/system';
+import { Theme } from 'web-components/lib/theme/muiTheme';
 
 /**
  * @returns a Material UI `Link` - will use React Router for local links
  */
-export const Link: React.FC<{ href: string; className?: string }> = ({
-  href,
-  children,
-  className,
-}) => {
-  const isInternalLink = (href: string): boolean => href.startsWith('/');
+export const Link: React.FC<{
+  href: string;
+  className?: string;
+  sx?: SxProps<Theme>;
+  target?: '_blank' | '_self';
+}> = ({ href, children, target, className, sx }) => {
+  const isInternalLink = (href: string): boolean =>
+    !!href && href.startsWith('/');
 
   return isInternalLink(href) ? (
-    <MuiLink className={className} component={RouterLink} to={href}>
+    <MuiLink
+      sx={sx}
+      className={className}
+      component={RouterLink}
+      to={href}
+      target={target}
+    >
       {children}
     </MuiLink>
   ) : (
-    <MuiLink className={className} href={href}>
+    <MuiLink sx={sx} className={className} href={href} target={target}>
       {children}
     </MuiLink>
   );
--- a/web-registry/src/lib/ecocredit.ts
+++ b/web-registry/src/lib/ecocredit.ts
@@ -12,6 +12,7 @@ import type {
   QueryBalanceResponse,
   BatchInfoWithBalance,
   QueryBatchesResponse,
+  QueryClassInfoResponse,
 } from '../types/ledger/ecocredit';
 import type { PageResponse } from '../types/ledger/base';
 
@@ -210,4 +211,17 @@ export const getTxsByEvent = (msgType: string): Promise<AxiosResponse> => {
       return qs.stringify(params);
     },
   });
+};
+
+export const getClassInfo = async (
+  class_id: string,
+): Promise<QueryClassInfoResponse> => {
+  try {
+    const { data } = await axios.get(
+      `${ledgerRestUri}/regen/ecocredit/v1alpha1/classes/${class_id}`,
+    );
+    return data;
+  } catch (err) {
+    throw new Error(`Error fetching class info: ${err}`);
+  }
 };
--- a/web-registry/src/lib/metadata-graph.ts
+++ b/web-registry/src/lib/metadata-graph.ts
@@ -1,0 +1,11 @@
+import axios, { AxiosResponse } from 'axios';
+
+export const getMetadata = async (hash: Uint8Array): Promise<AxiosResponse> => {
+  if (hash) {
+    // The on-chain metadata fields being stored as bytes for now, we need to decode it using base64 to get the corresponding IRI value as a string.
+    // This won't be needed once the on-chain metadata fields are converted to string: https://github.com/regen-network/regen-ledger/issues/829
+    const iri = atob(hash.toString());
+    return axios.get(`${process.env.REACT_APP_API_URI}/metadata-graph/${iri}`);
+  }
+  return Promise.reject('No metadata hash provided');
+};
--- a/web-registry/src/pages/CreditClassDetails.tsx
+++ b/web-registry/src/pages/CreditClassDetails.tsx
@@ -1,107 +1,28 @@
-import React, { useState } from 'react';
-import { makeStyles } from '@mui/styles';
-import { useParams, useNavigate, Routes, Route } from 'react-router-dom';
-import cx from 'clsx';
-
-import { Theme } from 'web-components/lib/theme/muiTheme';
-import Section from 'web-components/lib/components/section';
-import Description from 'web-components/lib/components/description';
-import Modal from 'web-components/lib/components/modal';
-import Banner from 'web-components/lib/components/banner';
-import MoreInfoForm from 'web-components/lib/components/form/MoreInfoForm';
-import { SwitchFooter } from 'web-components/lib/components/fixed-footer/SwitchFooter';
-import { BlockContent } from 'web-components/lib/components/block-content';
+import React, { useEffect, useState } from 'react';
+import { useParams, Routes, Route } from 'react-router-dom';
 
 import mock from '../mocks/mock.json';
-import { HeroTitle } from '../components/molecules';
-import {
-  ImpactSection,
-  ResourcesSection,
-  StepsSection,
-  MoreProjectsSection,
-  MediaSection,
-  // CreditClassConnectSection,
-  CreditClassOverviewSection,
-} from '../components/organisms';
-import hero from '../assets/credit-class-grasslands-hero.png';
-import getApiUri from '../lib/apiUri';
-import { onBtnClick } from '../lib/button';
-import { useMoreProjectsQuery } from '../generated/graphql';
 import { useAllCreditClassQuery } from '../generated/sanity-graphql';
+import { useCreditClassByOnChainIdQuery } from '../generated/graphql';
 import { client } from '../sanity';
+import { CreditClassDetailsWithContent } from './CreditClassDetailsWithContent';
+import { CreditClassDetailsSimple } from './CreditClassDetailsSimple';
+import { getClassInfo } from '../lib/ecocredit';
+import { getMetadata } from '../lib/metadata-graph';
+import { ClassInfo } from '../types/ledger/ecocredit';
 
 interface CreditDetailsProps {
   isLandSteward?: boolean;
 }
 
-const useStyles = makeStyles<Theme>((theme: Theme) => ({
-  root: {
-    display: 'flex',
-    flexDirection: 'column',
-    background: theme.palette.primary.main,
-  },
-  heroMain: {
-    maxWidth: theme.typography.pxToRem(744),
-    [theme.breakpoints.up('sm')]: {
-      paddingBottom: theme.spacing(22),
-    },
-    [theme.breakpoints.down('sm')]: {
-      paddingBottom: theme.spacing(12),
-    },
-  },
-  title: {
-    [theme.breakpoints.down('sm')]: {
-      fontSize: theme.typography.pxToRem(32),
-    },
-  },
-  titleWrap: {
-    paddingBottom: theme.spacing(8),
-  },
-  introSection: {
-    paddingBottom: theme.spacing(22.25),
-  },
-  overviewSection: {
-    paddingTop: 0,
-  },
-  sectionPadding: {
-    [theme.breakpoints.up('sm')]: {
-      paddingTop: theme.spacing(30),
-      paddingBottom: theme.spacing(30),
-    },
-    [theme.breakpoints.down('sm')]: {
-      paddingTop: theme.spacing(20),
-      paddingBottom: theme.spacing(20),
-    },
-  },
-  sectionDescription: {
-    [theme.breakpoints.up('sm')]: {
-      fontSize: theme.typography.pxToRem(22),
-      maxWidth: theme.typography.pxToRem(752),
-    },
-    [theme.breakpoints.down('sm')]: {
-      fontSize: theme.typography.pxToRem(18),
-    },
-  },
-  flex: {
-    display: 'flex',
-  },
-  stepsSection: {
-    background: 'transparent',
-    [theme.breakpoints.between('xs', 'lg')]: {
-      paddingLeft: theme.spacing(8),
-      paddingRight: theme.spacing(8),
-    },
-    [theme.breakpoints.down('sm')]: {
-      paddingLeft: theme.spacing(4),
-      paddingRight: theme.spacing(4),
-    },
-  },
-}));
-
 function CreditClassDetails(): JSX.Element {
   return (
     <Routes>
-      <Route path="/*" element={<CreditClassDetail isLandSteward={true} />} />
+      <Route
+        caseSensitive
+        path="/*"
+        element={<CreditClassDetail isLandSteward={true} />}
+      />
       <Route
         path="buyer"
         element={<CreditClassDetail isLandSteward={false} />}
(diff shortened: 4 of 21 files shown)


web-registry/src/lib/metadata-graph.ts
```diff
@@ -0,0 +1,7 @@
+import axios, { AxiosResponse } from 'axios';
+
+export const getMetadata = async (hash: string): Promise<AxiosResponse> => {
+  return axios.get(
+    `${process.env.REACT_APP_API_URI}/metadata-graph/regen:${hash}.rdf`,
```
blushi, line 5: the on chain metadata will store the full iri, not just the hash
```suggestion
    `${process.env.REACT_APP_API_URI}/metadata-graph/${iri}`,
```

web-registry/src/mocks/mocks.ts
```diff
@@ -10,6 +10,8 @@ export interface CreditClass {
   methodologyUrl: string;
   methodologyId: string;
   imgSrc: string;
+  admin?: string;
+  issuers?: string[];
```
blushi, line 14: why do you need this? the admin and issuers are available on chain: https://docs.regen.network/modules/ecocredit/02_state.html#class-info-table

web-registry/src/pages/CreditClassDetails.tsx
```diff
@@ -115,215 +31,55 @@ function CreditClassDetails(): JSX.Element {
...
-  const styles = useStyles();
-  const navigate = useNavigate();
-
   let { creditClassId } = useParams();
+  const [ledgerClass, setLedgerClass] = useState<any>(undefined);
+  const [metadata, setMetadata] = useState<any>(undefined);
 
   const { data } = useAllCreditClassQuery({ client });
+  const { data: dbData } = useCreditClassByOnChainIdQuery({
+    variables: { onChainId: creditClassId?.toUpperCase() || '' },
```
blushi, line 40: we should just have `credit_class/C01` for instance as path, no need to make it uppercase
URLs are case-sensitive

web-registry/src/pages/CreditClassDetails.tsx
```diff
@@ -115,215 +31,55 @@ function CreditClassDetails(): JSX.Element {
...
+  });
+
+  useEffect(() => {
+    if (creditClassId) {
+      getClassInfo(creditClassId.toUpperCase()).then(res => {
```
blushi, line 45: ditto

web-registry/src/mocks/mocks.ts
mhagel: I meant to delete this. Thx

web-registry/src/graphql/AllCreditClasses.graphql
```diff
@@ -4,11 +4,13 @@ query AllCreditClasses {
       id
       uri
       standard
+      onChainId
       creditClassVersionsById(orderBy: CREATED_AT_DESC, first: 1) {
         nodes {
           name
           description
           image
+          metadata
```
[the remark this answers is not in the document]
mhagel: Removed - this was also part of a draft version.

web-registry/src/lib/metadata-graph.ts
mhagel: Right now the ledger `metadata` field has to be `bytes`. As a proof of concept, I used base64 encoding. I know eventually we will use our custom `iri-gen` script. 

I guess for now, the best approach is to store the full IRI, base64 encoded, in the ledger `metadata`. Then when we fetch it here, decode it to get a more readable IRI, and fetch from `metadata-graph`?

mhagel: Another question. Something I don't get conceptually... In the `metadata-graph` table, it sounds like we plan to use the IRI as key, and the JSON-LD as value. 

If the key includes a hash of that JSON-LD, why do we need to store (and fetch) the data at all? Can't we just decode the hashed part of the IRI?

wgwz:
My understanding is that this is a one-way hash, in that it can't be 
reversed to get back the original data. And it just uniquely identifies 
the JSON document (the metadata). But I would love to know if I'm 
missing something here!

web-registry/src/pages/CreditClassDetails.tsx
mhagel: URLs typically aren't uppercase though. This would cover the case of a link that is shared or typed lowercase. Maybe it's an edge case, but seems common enough to me.

web-registry/src/lib/metadata-graph.ts
wgwz: My understanding is that this is a one-way hash, in that it can't be
reversed to get back the original data. And it just uniquely identifies
the JSON document (the metadata). But I would love to know if I'm
missing something here!

mhagel: That would explain it. Thanks @wgwz! 
I also realized there are the timestamps in the table, which can be very relevant.

web-registry/src/pages/CreditClassDetails.tsx
```diff
@@ -115,215 +31,55 @@ function CreditClassDetails(): JSX.Element {
...
+        const classInfo: any = res?.data;
```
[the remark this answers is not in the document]
mhagel: done

web-registry/src/pages/CreditClassDetails.tsx
mhagel: moved toUpperCase up, to avoid repeating

mhagel: see above

web-registry/src/lib/metadata-graph.ts
blushi: @wgwz is right.
we've got timestamps in the table just to be consistent with all the tables we have.

> I guess for now, the best approach is to store the full IRI, base64 encoded, in the ledger metadata. Then when we fetch it here, decode it to get a more readable IRI, and fetch from metadata-graph?

yes

web-registry/src/pages/CreditClassDetails.tsx
blushi: > URLs typically aren't uppercase though. 

https://schema.org/url and https://schema.org/URL are two different URLs for instance.

> And actually, it seems that react-router or the browser is not permitting an uppercase path - it resets it to lowercase.

there's a `sensitive` option for Route's in react-router: https://v5.reactrouter.com/core/api/Route/sensitive-bool
I think we should keep the case-sensitiveness as this might be confusing for users to know whether the class id is `C01` or `c01`.

mhagel: Sounds good. It is called `caseSensitive` now. Updated

web-registry/src/pages/CreditClassDetailsSimple.tsx
```diff
@@ -0,0 +1,313 @@
...
+            <Description
+              className={styles.description}
+              key={firstMethodology?.['schema:name']}
+            >
+              {firstMethodology?.['schema:name']}
+            </Description>
+            {count > 1 && (
+              <Link
+                sx={{ display: 'flex', alignItems: 'center' }}
+                href="TODO"
```
mhagel, line 131: This URL is not defined in the schema standard yet. I have a pending question about how it will show up here, but easy to fix once it is decided

web-registry/src/lib/metadata-graph.ts
mhagel: Done. Left a TODO for clarification,

web-registry/src/lib/metadata-graph.ts
```diff
@@ -0,0 +1,11 @@
+import axios, { AxiosResponse } from 'axios';
+
+export const getMetadata = async (hash: Uint8Array): Promise<AxiosResponse> => {
+  if (hash) {
+    // TODO: this is a POC for getting bytes from the ledger and decoding to fetch from DB.
+    // In the future, we will use our custom iri-gen script to manage the hashes, not atob().
```
blushi, line 6: I don't think this comment is relevant.
The iri-gen script doesn't change anything to that, it just gives you the hashed iri for some JSON-LD data, that you can then use in the on chain `metadata` field
But since this metadata field is still bytes, the string will need to be base64 encoded in order to store it on-chain, that's it.

Maybe we should just replace it with the following:
```suggestion
The on-chain metadata fields being stored as bytes for now, we need to decode it using base64 to get the corresponding IRI value as a string.
This won't be needed once the on-chain metadata fields are converted to string: https://github.com/regen-network/regen-ledger/issues/829
```

web-registry/src/pages/CreditClassDetailsSimple.tsx
```diff
@@ -0,0 +1,313 @@
...
+                      <Box sx={{ display: 'flex', alignItems: 'center' }}>
+                        <Description
+                          sx={{ mr: 1 }}
+                          className={styles.description}
+                        >
+                          {metadata?.['regen:sourceRegistry']?.['schema:name']}
+                        </Description>
+                        <Link
+                          href={
+                            metadata?.['regen:sourceRegistry']?.['schema:URL']
```
blushi, line 223:
```suggestion
                            metadata?.['regen:sourceRegistry']?.['schema:url']
```

[schema:URL](https://schema.org/URL) is a datatype, we need the property here: [schema:url](https://schema.org/url)

blushi (changes requested):

web-registry/src/pages/CreditClassDetailsSimple.tsx
```diff
@@ -0,0 +1,313 @@
...
+          width: { xs: '100%', sm: '50%' },
+          pr: 4,
+        }}
+        className={styles.marginBottom}
+      >
+        <Label className={styles.label}>{label}</Label>
+        <Description className={styles.description}>{data}</Description>
+      </Box>
+    ) : (
+      <></>
```
haveanicedavid, line 164: Are you returning a fragment here because of typescript complaining if you return `null`?

Both work and there's not much of a perf difference, but as an FYI this is part of why I prefer `React.FC` instead of manually typed functions like this (typescript doesn't complain when you return `null`). 

It seems most are recommending away from `React.FC` because of the implicit types it adds (it'll always add a `children` type, so TS won't error if you try to use children in a place it won't be rendered), but IMO those are really minor problems and it helps avoid writing boilerplate (for `children` type when you use it, and for the return type)

I also prefer the [return early](https://forum.freecodecamp.org/t/the-return-early-pattern-explained-with-javascript-examples/19364) pattern, as it avoids deep nesting... something like 
```tsx
if (!data) return null

return (
  <Box> {...} </Box>
)
```

web-registry/src/pages/CreditClassDetailsWithContent.tsx
```diff
@@ -0,0 +1,322 @@
...
+              preTitle: Maybe<string> | undefined;
+              descriptionRaw: any;
+              stepCards: Maybe<Maybe<StepCardFieldsFragment>[]> | undefined;
+            },
+            index: React.Key | null | undefined,
+          ) => (
+            <div
+              className={cx('topo-background-alternate', styles.flex)}
+              key={index}
+            >
```
haveanicedavid, line 256:
```suggestion
            <Box key={index} sx={{ display: 'flex' }} className="topo-background-alternate">
```
I mentioned this in [victor's PR](https://github.com/regen-network/regen-web/pull/856#discussion_r819893049), but we should probably avoid using `makeStyles` as much as possible moving forward, given it probably won't work in react 18

web-registry/src/lib/ecocredit.ts
```diff
@@ -130,3 +131,11 @@ const getAccountEcocreditsForBatch = (
...
   );
 };
+
+export const getClassInfo = async (
+  class_id: string,
+): Promise<AxiosResponse<QueryClassInfoResponse>> => {
+  return axios.get(
+    `${ledgerRestUri}/regen/ecocredit/v1alpha1/classes/${class_id}`,
+  );
+};
```
haveanicedavid, line 141: Minor nit: you don't need `async` for this function as-is

I thought through this a bit and figured for our use, it makes sense to `await` the response data and return within this function (so you don't have to dig through `response.data` every time you use) +  a basic trycatch for error handling. Something like 

```suggestion
export const getClassInfo = async (
  class_id: string,
): Promise<QueryClassInfoResponse> => {
  try {
    const { data } = await axios.get(
      `${ledgerRestUri}/regen/ecocredit/v1alpha1/classes/${class_id}`,
    );
    return data;
  } catch (err) {
    throw new Error(`Error fetching class info: ${err}`);
  }
};
```

haveanicedavid (commented):
A few nits but overall looks good so far!

web-registry/src/components/atoms/Link.tsx
```diff
@@ -1,23 +1,32 @@
...
-export const Link: React.FC<{ href: string; className?: string }> = ({
-  href,
-  children,
-  className,
-}) => {
-  const isInternalLink = (href: string): boolean => href.startsWith('/');
+export const Link: React.FC<{
+  href: string;
+  className?: string;
+  sx?: SxProps;
```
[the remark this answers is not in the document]
mhagel: thanks. done.

web-registry/src/pages/CreditClassDetails.tsx
```diff
@@ -115,215 +37,61 @@ function CreditClassDetails(): JSX.Element {
...
+
+  const { data: contentData } = useAllCreditClassQuery({ client });
+  const { data: dbData } = useCreditClassByOnChainIdQuery({
+    variables: { onChainId: creditClassId || '' },
+  });
+
+  useEffect(() => {
+    if (creditClassId) {
+      getClassInfo(creditClassId).then(
+        (res: AxiosResponse<QueryClassInfoResponse>) => {
```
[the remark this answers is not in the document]
mhagel: good call

web-registry/src/pages/CreditClassDetailsSimple.tsx
mhagel: thanks. done.

web-registry/src/pages/CreditClassDetailsSimple.tsx
mhagel: updated to WIP standard, but this may still need to be updated.

web-registry/src/pages/CreditClassDetailsSimple.tsx
```diff
@@ -0,0 +1,313 @@
...
+          title="Projects"
+          projects={projects}
+        />
+      </div>
+    ) : (
+      <></>
+    );
+  };
+
+  const ApprovedMethodologies: React.FC<{ methodologies: any[] }> = ({
```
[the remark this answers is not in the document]
mhagel: It is a SHACL-style interface. I added it to the types, but it is subject to change, and will need to be imported eventually (it is a WIP).

web-registry/src/pages/CreditClassDetailsSimple.tsx
mhagel: yeah that's better, I must have gotten lazy here. Thanks.

web-registry/src/types/ledger/ecocredit.ts
```diff
@@ -50,3 +50,35 @@ export interface QueryBasketsResponse {
...
+   */
+  metadata: Uint8Array;
+  credit_type: CreditType;
+}
+
+export interface QueryClassInfoResponse {
+  info?: ClassInfo;
+}
+
+export interface CreditType {
```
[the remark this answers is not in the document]
mhagel: done

web-registry/src/pages/CreditClassDetailsWithContent.tsx
mhagel: I'm in agreement, but.... 
This is probably a sign that my code isn't modular enough, but I haven't figured out how to re-use styles with `styled` without a massive refactor of all the components it's touching. 

In this case, I mostly just renamed this file to separate the "content" credit class pages from the "simple," so this part is copy-paste.

In other cases, it is like unscrambling an omelette. But I will try to use `styled` moving forward.

haveanicedavid: copy-paste makes sense! There will definitely be places we need to `makeStyles` to pass some custom css to components which accept `classes` or `className` as props. 

I don't think `styled` is a great replacement for this sort of thing either either... it's good for components moving forward, but requires some different reasoning for passing styles via props. We could also just use `sx` for everything and avoid `styled` for most... there's a small perf hit compared to `styled` or passing a css class, but it's minor.  

I've been playing with some ideas, and it seems that `styled` + `sx` (which you can pass via props and combine) works great, and `makeStyles` + `sx` generally works well (sometimes passed `sx` props will be overridden by the `makeStyles` class), but combining the 3 can get a bit wonky. Sometimes it works perfectly, other times not so much, and i'm not entirely sure why. My current plan is to use `styled` for base styles (anything that won't be modified through different uses cases), `sx` for anything that might change, and handling modifications through a passed `sx` prop as [mui suggests](https://mui.com/system/the-sx-prop/#passing-sx-prop)

That's all a tangent though. TLDR: no worries at all and no need to change anything, just wanted to make sure you're aware. I think everyone is at this point though, so i'll stop reminding everyone 😀

haveanicedavid (approved):
Right now the deploy preview isn't showing much: 
image

and there's a call to `https://regen-registry-server-staging.herokuapp.com/metadata-graph/REDD+` that's erroring (`Invalid IRI, it must be of the form regen:<iri-hash>.rdf`), but code looks good and I assume those issues are happening because of the hambach / redwood situation. 

Pre-approving - nice work!

blushi:
> and there's a call to https://regen-registry-server-staging.herokuapp.com/metadata-graph/REDD+ that's erroring (Invalid IRI, it must be of the form regen:<iri-hash>.rdf),


@haveanicedavid this is coming from the server which only accepts IRI with the given form: https://github.com/regen-network/registry-server/blob/master/server/routes/metadata-graph.ts#L12-L13

web-registry/src/pages/CreditClassDetailsWithContent.tsx
victordelval: I've created a discussion on GitHub to formalize this topic with some priority. I have also dedicated a space to it in our Dev Docs in Notion..

[Styling strategies for MUI v3 to v5](https://github.com/regen-network/regen-web/discussions/863)

victordelval (approved):

victordelval:
Super! :+1:

PR #866, opened 2022-03-14
David/825 credit batch details

haveanicedavid:
## Description

Closes: regen-network/regen-registry#825

I'm having a hard time with the metadata fetching right now - I think things are mostly configured correctly, and there's a value in the `metadata` field from the batches call I use to hit the proxied metadata endpoint, but it's erroring.

There are still some tweaks to the data needed I think, but most of the code should be ready for review

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

- [ ] reviewed tests


web-components/src/components/section/index.tsx
```diff
@@ -31,32 +33,36 @@ interface StyleProps {
   titleAlign?: 'left' | 'right' | 'inherit' | 'center' | 'justify' | undefined;
 }
 
+const Root = styled(Box, {
```
haveanicedavid, line 36: convert root `section` element to use `styled` so passed `sx` behaves better

web-components/src/components/text-layouts/LabeledDetail.tsx
```diff
@@ -0,0 +1,31 @@
...
+  },
+}));
+
+/** Grey label over child elements */
+export const LabeledDetail: React.FC<{
+  label: string;
+  sx?: SxProps<Theme>;
+  styles?: {
+    label?: SxProps<Theme>;
+  };
```
haveanicedavid, line 25: this is a pattern that makes the most sense to me, but i'm not sure about it. Basically the `sx` prop would always be passed to the root element, then any sub-styles would be handled via a `styles` object. 

The alternative would be to have something like `styles.root` - but it's harder to combine `sx` so I think it makes sense to do what i've done here

blushi (commented):
looking good, just a couple of small comments

web-components/src/components/section/index.tsx
```diff
@@ -31,32 +33,36 @@ interface StyleProps {
...
+  name: 'RegenSection',
+  shouldForwardProp: prop => prop !== 'sx' && prop !== 'withSlider',
```
haveanicedavid, line 38: because there's no `useStyles` to pass props to, they need to be passed to the component directly. `shouldForwardProps` is needed to not pass them to underlaying component

victordelval (approved):
Without wanting to bring any changes right now, I think the `BatchDetails` page component should be organized differently, be much smaller, as a thin composition of the big semantic subcomponents (no styling here), and instead elsewhere such subcomponents (if they are in the same file, I think it is better that they are below the main component.. and the main component exported as default..)

In any case, as a general rule, we should be creating smaller components, decoupling responsibilities and reusing more code.

We will talk about these things next month, after the deliveries we have now.

haveanicedavid:
@victordelval  Agree we should err on the side of smaller components which is actually why this is split into several sub-components - I originally had them in the same file for convenience, meaning to pull them out later but then hesitated because I wasn't sure how reusable these components would really be. 

Looking back on the code though, it doesn't make sense to have that file so big. Just updated the PR extracting those out

I avoid export default as much as possible when working with TS, but actually prefer having `export`s at the top of the file... I put the other functions at top to try and follow conventions i'd seen in the codebase. 

> In any case, as a general rule, we should be creating smaller components, decoupling responsibilities and reusing more code.

Totally agree. I do think it'll take a bit of planning though, and appreciate you taking the charge on tackling!

haveanicedavid:
I should also add a general note RE convos with @blushi : the NCT story took longer than I planned, and I wasn't able to issue a new batch on chain for this one. If we need to get this merged ASAP, someone else might have to take on that task. Otherwise, I can tackle when I am back in town on the 29th

web-registry/src/pages/BatchDetails.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  const { data: offchainData } = useBatchDetailsQuery({
+    skip: !batchDenom,
+    variables: { batchDenom: batchDenom as string },
+  });
+  const projectHandle =
+    offchainData?.creditVintageByBatchDenom?.projectByProjectId?.handle || '';
+
+  if (!batch) return null;
+  // TODO should there be a not found component or redirect?
```
blushi, line 52: cc/ @erikalogie we generally don't have such thing implemented (for projects, credit classes, and here credit batches), maybe something to think about, either we could just display the existing "not found" page or some new component

web-registry/src/pages/BatchDetails.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const [metadata, setMetadata] = useState<BatchMetadataLD>();
+
+  useEffect(() => {
+    const fetch = async (): Promise<void> => {
+      if (batchDenom) {
+        try {
+          const batch = await getBatchWithSupplyForDenom(batchDenom);
+          setBatch(batch);
+          if (batch.metadata) {
+            const { data: metadata } = await getMetadata(batch.metadata);
```
blushi, line 33:
```suggestion
            const { metadata } = await getMetadata(batch.metadata);
```

I've tested this with https://deploy-preview-866--regen-registry.netlify.app/credit-batches/C01-20170101-20180101-006, a credit batch that I've just created with some metadata but somehow the metadata is not getting displayed while it's indeed returned as you can see on:
image
The code suggestion above should fix it.

blushi (changes requested):

blushi (commented):
Somehow the report links in the metadata card are white:
image
image

web-registry/src/pages/BatchDetails.tsx
```diff
@@ -0,0 +1,110 @@
...
+            gap: 4,
+            flexWrap: 'wrap',
+            justifyContent: 'space-between',
+            alignItems: 'center',
+          }}
+        >
+          <Title variant="h2">Credit Batch Details</Title>
+          <OutlinedButton sx={{ maxWidth: 'max-content', px: 7 }} size="small">
+            view in portfolio
+          </OutlinedButton>
```
blushi, line 70: I guess we should only display this if the current user does have some of these ecocredits
Also it's missing the actual link

blushi (changes requested):

erikalogie:
@blushi I think we can display the "not found" page for now, I'll add a design task to think about this though as I suppose the development should be considered a separate task too?

web-registry/src/pages/BatchDetails.tsx
haveanicedavid: I completely overlooked that - good catch. Updated

haveanicedavid:
I've just updated this PR with several things:
1. font color should display correctly on the metadata items
2. using @mhagel 's `LinkWithArrow` component
3. Only displays the 'view in portfolio' button (which links to 'my ecocredits') if the batch is in current user's batches
4. Added a basic `<Loading />` component to `web-components`. There are several areas I think this could make the UI a bit cleaner on initial load

haveanicedavid:
@blushi I spent time today trying to get a credit batch on staging associated with my account I can view based on the call last week, but didn't get it so am . Would you mind checking if this looks correct?

blushi:
> @blushi I think we can display the "not found" page for now, I'll add a design task to think about this though as I suppose the development should be considered a separate task too?

indeed
cc/ @haveanicedavid

blushi:
> @blushi I spent time today trying to get a credit batch on staging associated with my account I can view based on the call last week, but didn't get it so am . Would you mind checking if this looks correct?

here's one batch you have credits for:
http://redwood.regen.network:1317/regen/ecocredit/v1alpha1/batches/C01-20180101-20200101-001/balance/regen1rhvx7kpxtd0em0t2u87xu5a3xc9gs7evvr5v29
(using the address you once shared with me on slack)

web-components/src/components/icons/icons.stories.tsx
```diff
@@ -126,6 +127,9 @@ export const linkedInIcon = (): JSX.Element => <LinkedInIcon />;
 
 export const linkIcon = (): JSX.Element => <LinkIcon color="black" />;
 
+/** this isn't really an icon, but it makes sense here I think in storybook */
+export const loadingIcon = (): JSX.Element => <Loading />;
+
```
blushi, line 132: this is a bit weird to put that inside icons.stories.tsx while the component live somewhere else
why not just having it in icons/ folder?

web-registry/src/pages/BatchDetails.tsx
```diff
@@ -0,0 +1,130 @@
...
+  BatchMetadata,
+  BatchTotalsGrid,
+} from '../components/molecules';
+import { NotFoundPage } from './NotFound';
+import { useWallet } from '../lib/wallet';
+import { useEcocredits } from '../hooks';
+
+export const BatchDetails: React.FC = () => {
+  const { batchDenom } = useParams();
+  const [loading, setLoading] = useState(true);
```
blushi, line 27: maybe that should be false by default

web-registry/src/pages/BatchDetails.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const { data: offchainData } = useBatchDetailsQuery({
+    skip: !batchDenom,
+    variables: { batchDenom: batchDenom as string },
+  });
+  const projectHandle =
+    offchainData?.creditVintageByBatchDenom?.projectByProjectId?.handle || '';
+
+  // TODO placeholder until we have a more descriptive not found graphic
+  if (loading) return <Loading />;
+  if (!batch) return <NotFoundPage />;
```
blushi, line 66:
```suggestion
  if (loading) return <Loading />;
  
  // TODO placeholder until we have a more descriptive not found graphic
  if (!batch) return <NotFoundPage />;
```

web-registry/src/pages/BatchDetails.tsx
```diff
@@ -0,0 +1,130 @@
...
+          console.error(err); // eslint-disable-line no-console
+        } finally {
+          setLoading(false);
+        }
+      }
+    };
+    fetch();
+  }, [batchDenom]);
+
+  const { data: offchainData } = useBatchDetailsQuery({
```
blushi, line 57: shouldn't we also pull the loading state from this query to know whether we display <Loading /> or not?

haveanicedavid: Yes, good idea! I threw it in last minute - will update

web-components/src/components/icons/icons.stories.tsx
haveanicedavid: I moved this to a `loading.stories.tsx` inside of the `/loading` folder - given the talk about using skeleton components, I figure they could live under this folder as well

blushi (approved):

web-registry/src/components/molecules/BatchTotalsGrid.tsx
```diff
@@ -0,0 +1,40 @@
...
+    container
+    rowGap={6}
+    columnGap={2}
+    sx={{ justifyContent: 'space-between', ...sx }}
+  >
+    <GridItem>
+      <LabeledNumber label="Total Credits Issued" number={batch.total_amount} />
+    </GridItem>
+    <GridItem>
+      <LabeledNumber label="Credits Tradable" number={batch.total_amount} />
```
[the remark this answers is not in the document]
blushi: see how this is handled on the homepage credit batches table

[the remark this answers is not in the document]
haveanicedavid: just corrected - it now references `batch.tradable_supply` like the credit batches table. Sorry about that!

web-components/src/components/text-layouts/LabeledDetail.tsx
```diff
@@ -0,0 +1,31 @@
...
+    fontSize: 14,
+  },
+}));
+
+/** Grey label over child elements */
+export const LabeledDetail: React.FC<{
+  label: string;
+  sx?: {
+    root?: SxProps<Theme>;
+    label?: SxProps<Theme>;
```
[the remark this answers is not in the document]
mhagel: Looks like this is the way to go.

web-registry/src/components/molecules/BatchInfoGrid.tsx
```diff
@@ -0,0 +1,64 @@
...
+  sx?: SxProps<Theme>;
+}> = ({ batch, projectHandle: project, sx }) => (
+  <Grid
+    container
+    rowGap={6}
+    columnGap={2}
+    sx={{ justifyContent: 'space-between', ...sx }}
+  >
+    <GridItem>
+      <BatchDetail label="Batch Denom">{batch.batch_denom}</BatchDetail>
```
mhagel, line 22: Good name BatchDetail. I could only think of `LineItem`.  :)

mhagel (approved):
nice work

blushi (approved):

haveanicedavid:

PR #868, opened 2022-03-17
Update project metadata

mhagel:
## Description

Closes: regen-network/regen-registry#805

Adds new metadata fields to Project page.

Main changes are in `ProjectTopSection` and `AdditionalMetaData`

Viewable at:
https://deploy-preview-868--regen-registry.netlify.app/projects/123

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

- [ ] reviewed tests


web-registry/src/components/organisms/EntityDisplayForm.tsx
```diff
@@ -46,6 +46,7 @@ export interface EntityDisplayValues {
   'http://regen.network/landSteward'?: DisplayValues;
   'http://regen.network/projectDeveloper'?: DisplayValues;
   'http://regen.network/projectOriginator'?: DisplayValues;
+  'regen:projectDeveloper'?: DisplayValues;
```
mhagel, line 49: @blushi I noticed the new standard uses this format `'regen:projectDeveloper'`, although we were using `'http://regen.network/projectDeveloper'` before. How do you think we should handle this right now?
 I was hoping to avoid increasing the scope and redoing all of the form input properties, but maybe it is unavoidable? I would also have to make sure all the existing project metadata was updated in PROD.

web-components/src/components/buttons/ExpandButton.tsx
```diff
@@ -0,0 +1,45 @@
...
+      }}
+      endIcon={
+        expanded ? (
+          <ArrowDownIcon direction="up" color={theme.palette.secondary.main} />
+        ) : (
+          <ArrowDownIcon
+            direction="down"
+            color={theme.palette.secondary.main}
+          />
+        )
```
haveanicedavid, line 35: Minor nit: this can be DRYd up a bit
```suggestion
          <ArrowDownIcon
            direction={expanded ? "up" : "down"}
            color={theme.palette.secondary.main}
          />
```

(note: i'm having a hard time with this suggestion - it's pseudo codey)

web-registry/src/components/molecules/AdditionalMetadata.tsx
```diff
@@ -0,0 +1,106 @@
+import React, { useState } from 'react';
```
haveanicedavid, line 1: might want to name this `AdditionalProjectMetadata`

web-registry/src/components/molecules/LineItemLabelAbove.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+import Description from 'web-components/lib/components/description';
+import { Label } from 'web-components/lib/components/label';
+
+interface LineItemProps {
+  label: string;
+  data: string | JSX.Element | number;
+}
+
+export const LineItemLabelAbove: React.FC<LineItemProps> = ({
```
haveanicedavid, line 12: looks like we might have some duplicate styles between this PR and my two open PRs. I don't imagine it'll be a big issue, but could be worth circling back on when we merge (depending on which gets merged first)

web-registry/src/types/project/project.ts
```diff
@@ -0,0 +1,51 @@
+export type VcsProjectMetadata = {
```
haveanicedavid, line 1: you already have this built so not sure this would be helpful for you, but for my story I ran the JSON structures through a basic type generator and used them. Just makes it so you don't have to modify as much (moving forward, not suggesting you change this). see: https://github.com/regen-network/regen-web/pull/866/files#diff-557a5b59cd7201c768fefa749405991354b999f4b85d50d79e9e00f1efda2043

web-registry/src/components/molecules/AdditionalMetadata.tsx
```diff
@@ -0,0 +1,106 @@
...
+  additionalCertification: any;
+}
+
+const AdditionalMetadata: React.FC<MetadataProps> = ({
+  metadata,
+  creditClass,
+  creditClassVersion,
+  additionalCertification,
+}) => {
+  const theme = useTheme<Theme>();
```
haveanicedavid, line 26: I believe if you import `@mui/material` (instead of `@mui/styles`), you shouldn't have to pass the theme type

haveanicedavid (approved):
Looks good to me! This involved a lot of tedious data - great work. I haven't been able to test too much manually, but code looks good minus a few nits so i'm pre-approving. 

Know you didn't do it in this story, but I love the slide animation btw

web-registry/src/components/organisms/EntityDisplayForm.tsx
blushi: these are the same things, the only difference is that `regen:projectDeveloper` corresponds to the compacted JSON-LD document form where we define "regen": "http://regen.network/" in the `@context` (see docs https://json-ld.org/spec/latest/json-ld/#compacted-document-form) 
I guess for now let's just use the expanded version in the metadata (eg `'http://regen.network/projectDeveloper'`) so that we don't have to touch this for now.
Later on, we might think of using https://github.com/digitalbazaar/jsonld.js/ so we can just work from same document form in the code without necessarily changing the underlying data (we do already have an issue to track that: https://github.com/regen-network/regen-registry/issues/724)

mhagel: Thanks. Using expanded format for now.

web-components/src/components/buttons/ExpandButton.tsx
mhagel: done

web-registry/src/types/project/project.ts
mhagel: Looks good. In this case, I have to temporarily deviate from the standard, so this is a helper for now. I will leave a TODO to update once the standard is finished.

web-registry/src/components/molecules/AdditionalMetadata.tsx
mhagel: done

web-registry/src/components/organisms/ProjectTopSection.tsx
```diff
@@ -254,23 +258,17 @@ function ProjectTopSection({
     title: s.title || '',
     imageUrl: getSanityImgSrc(s.image),
   }));
-
   return (
     <Section classes={{ root: styles.section }}>
       <Grid container>
         <Grid item xs={12} md={8} sx={{ pr: { md: 19 } }}>
-          {/* TODO Show on-chain project id if no off-chain name */}
```
blushi, line 262: why removing this? this should be kept here

web-registry/src/components/organisms/ProjectTopSection.tsx
```diff
@@ -372,6 +345,7 @@ function ProjectTopSection({
               {landStory}
             </Description>
           )}
+          {isVcsProject && <AdditionalProjectMetadata metadata={metadata} />}
```
blushi, line 348: I believe we should show this AdditionalProjectMetadata if at least one of the metadata field that we need to show there is present, not just the vcs project id
the side effect is that right now, the project activity is not appearing anymore on the kasigau page, while it should now appear in the additional metadata section

web-registry/src/components/templates/ProjectDetails.tsx
```diff
@@ -125,34 +126,35 @@ function ProjectDetails(): JSX.Element {
...
-          geojson = togeojson.kml(dom);
-        } else {
-          geojson = JSON.parse(text);
-        }
-        setGeojson(geojson);
-      });
-  }
+  useEffect(() => {
+    if (!geojson && isGISFile) {
+      fetch(mapFile)
```
blushi, line 144: if there's no map file, we should use the `regen:location` geojson value instead

web-registry/src/components/organisms/ProjectTopSection.tsx
```diff
@@ -290,50 +288,24 @@ function ProjectTopSection({
...
-                    name={
-                      creditClassVersion.metadata?.[
-                        'http://regen.network/offsetGenerationMethod'
-                      ]
-                    }
-                  />
                 </>
               )}
-              {methodologyVersion && (
+              {!isVcsProject && (
```
blushi, line 293: we should just move the "OFFSET GENERATION METHOD" to the additional metadata section in any case (even on existing pages like wilmot or kasigau)

web-registry/src/types/project/vcs-project.ts
```diff
@@ -0,0 +1,69 @@
...
+// https://github.com/regen-network/regen-registry-standards/blob/a2d81baebe7756226a58b126378d1890ad4cadcc/jsonld/projects/C01-verified-carbon-standard-project.json
+
+export interface VcsProjectMetadataLD {
+  '@context': {
+    schema: 'http://schema.org/';
+    regen: 'http://regen.network/';
+    qudt: 'http://qudt.org/schema/qudt';
+    unit: 'http://qudt.org/2.1/vocab/unit';
+    xsd: 'http://www.w3.org/2001/XMLSchema#';
+  };
```
blushi, line 11: this context is useless here since you're using full IRIs in the rest of the data, except for qudt and xsd, which should be fixed

blushi (changes requested):
image
 this should link to the credit class page instead, eg /credit-classes/C02.


Also it seems like you haven't changed the project developer (and other project stakeholders) to use the metadata info (instead of the `party` info)

victordelval (approved):
Good job Mark!

web-registry/src/components/organisms/ProjectTopSection.tsx
mhagel: Ok. I was showing the VCS id when I deleted this, which is not even the same thing. Restored todo.

web-registry/src/components/templates/ProjectDetails.tsx
mhagel: done

web-registry/src/types/project/vcs-project.ts
mhagel: removed and fixed

web-registry/src/components/organisms/ProjectTopSection.tsx
mhagel: I added a VCS project ID so metadata will show for Kasigau. Still a question about whether to handle non-VCS metadata the same way.

web-registry/src/components/organisms/ProjectTopSection.tsx
blushi: Resolving this for now per our discussion with @erikalogie yesterday, since we only want to move this to the additional metadata section for VCS projects

web-registry/src/components/organisms/ProjectTopSection.tsx
blushi: sounds good

web-registry/src/components/templates/ProjectDetails.tsx
blushi: could you remove the http://regen.network/boundaries field from the "123" project on staging so we can make sure it works as expected?

blushi:
> Also it seems like you haven't changed the project developer (and other project stakeholders) to use the metadata info (instead of the party info)

It doesn't seem like this has been addressed yet.

erikalogie:
Should we remove the location of the project developer / issuer or have that as part of a follow up task since we won't have that in the metadata for VCS projects (and likely will eliminate from all projects moving forward)?

blushi:
> Should we remove the location of the project developer / issuer or have that as part of a follow up task since we won't have that in the metadata for VCS projects (and likely will eliminate from all projects moving forward)?

The location is already set as optional anyway. If there's no value for it, then nothing is displayed.

erikalogie:
Ok, that works!

mhagel:
@blushi 
>Also it seems like you haven't changed the project developer (and other project stakeholders) to use the metadata info (instead of the party info)

>It doesn't seem like this has been addressed yet.

Yes, I still planned to tackle this, but haven't yet. 

>maybe keeping the current pattern is actually best, ie having this info stored both in the party table and in the project metadata. When creating new projects, we'll just need to make sure we create the corresponding party/ies and org/s.

I'm not sure which way to go. I think I will continue with reading stakeholders from metadata. We could check for a party first, and fall back to metadata perhaps? 

That seems like it would make things easier for creating projects (?). That way there is no requirement to create new parties when creating projects. However, our existing UI does this I believe, so maybe it's not much of a burden?

web-registry/src/lib/transform.ts
```diff
@@ -181,6 +181,23 @@ export function getParty(
   };
 }
 
+const getPartyFromMetadata = (
```
mhagel, line 184: @blushi Here is a simple fallback to pull from metadata

web-registry/src/lib/ledger.ts
```diff
@@ -8,8 +8,8 @@ export const expLedger = process.env.REACT_APP_EXP_LEDGER;
 const expUriBase = 'exp-';
 const uriBase = `${getApiUri()}/${expLedger ? expUriBase : ''}`;
 
-// Simple proxy endpoint for REST requests. We check for chainId as an on/off switch.
-export const ledgerRESTUri = chainId ? `${uriBase}ledger-rest` : undefined;
```
mhagel, line 12: @blushi I removed this check because there are cases (like the new credit class page) where ledger-rest requests are not experimental, as I mentioned in slack. https://regen-network.slack.com/archives/C01LX9E8QN8/p1648155852710709?thread_ts=1647993852.056959&cid=C01LX9E8QN8

web-registry/src/components/organisms/ProjectTopSection.tsx
```diff
@@ -280,60 +281,36 @@ function ProjectTopSection({
...
                     url={
-                      creditClassVersion.metadata?.['http://schema.org/url']?.[
-                        '@value'
-                      ]
+                      isVcsProject
+                        ? `/credit-classes/${creditClass?.onChainId}`
+                        : creditClassVersion.metadata?.[
+                            'http://schema.org/url'
+                          ]?.['@value']
                     }
```
mhagel, line 289: @blushi Linking to new ledger credit class page now. But still linking to carbonplus-grasslands if not VCS. Not sure if this is the best logic.

blushi: I believe that'll work for now

web-registry/src/lib/ledger.ts
blushi: sounds good to me, although "experimental" is not the right word here (since experimental is meant to refer to features on hambach)

blushi:
> We could check for a party first, and fall back to metadata perhaps?

Sounds good to me!

web-registry/src/lib/transform.ts
```diff
@@ -181,6 +181,23 @@ export function getParty(
...
+  metadata: EntityDisplayValues,
+  role: EntityFieldName,
+): Party | undefined => {
+  const metadataRole = metadata[role];
+  if (!metadataRole) return undefined;
+
+  return {
+    name: metadataRole?.['http://schema.org/name'] || '',
+    description: metadataRole?.['http://schema.org/description'] || '',
+    type: 'ORGANIZATION', // to provide default image
```
blushi, line 194: this can be inferred from `metadataRole?.['@type']` instead

web-registry/src/lib/transform.ts
```diff
@@ -189,7 +206,9 @@ export function getDisplayParty(
   const showOnProjectPage =
     metadata?.[role]?.['http://regen.network/showOnProjectPage'];
   if (showOnProjectPage) {
-    return getParty(party);
+    const dbParty = getParty(party);
+    if (dbParty) return dbParty;
+    return getPartyFromMetadata(metadata, role);
```
blushi, line 211: adding a small comment describing why we use this logic here would be nice

web-registry/src/types/project/vcs-project.ts
```diff
@@ -0,0 +1,62 @@
...
+  'http://regen.network/vcsProjectId': {
+    '@type': 'xsd:unsignedInt';
+    '@value': number;
+  };
+  'http://regen.network/vcsProjectPage': {
+    '@type': 'http://schema.org/URL';
+    '@value': string;
+  };
+  'http://regen.network/projectDeveloper': {
+    '@type': 'http://regen.network/Organization';
```
blushi, line 15:
```suggestion
    '@type': 'http://regen.network/OrganizationDisplay';
```

web-registry/src/types/project/vcs-project.ts
```diff
@@ -0,0 +1,62 @@
...
+    'http://schema.org/name': string;
+    'http://schema.org/description': string;
```
blushi, line 17:
```suggestion
    'http://schema.org/description': string;
    'http://regen.network/showOnProjectPage': boolean;
```

web-registry/src/types/project/vcs-project.ts
```diff
@@ -0,0 +1,62 @@
...
+  'http://regen.network/vcsMethodology': {
+    'http://schema.org/name': string;
+    'http://schema.org/url': {
+      '@type': 'http://schema.org/URL';
+      '@value': string;
+    };
+  };
+  'http://regen.network/additionalCertification': {
+    'http://schema.org/name': string;
+    'http://schema.org/version': string;
```
blushi, line 38:
```suggestion
```

web-registry/src/types/project/vcs-project.ts
```diff
@@ -0,0 +1,62 @@
...
+    'http://schema.org/url': {
+      '@type': 'http://schema.org/URL';
+      '@value': string;
+    };
+  };
+  'http://regen.network/projectSize': {
+    'http://qudt.org/schema/qudt/#unit': {
```
blushi, line 45:
```suggestion
    'http://qudt.org/schema/qudt/unit': {
```

web-registry/src/components/organisms/ProjectTopSection.tsx
```diff
@@ -222,23 +218,33 @@ function ProjectTopSection({
...
+    ]?.['@value'] ||
+    metadata?.['http://regen.network/projectSize']?.[
+      'http://qudt.org/schema/qudt/numericValue'
+    ]?.['@value'];
   const unit: qudtUnit | undefined =
     metadata?.['http://regen.network/size']?.[
       'http://qudt.org/1.1/schema/qudt#unit'
+    ]?.['@value'] ||
+    metadata?.['http://regen.network/projectSize']?.[
+      'http://qudt.org/schema/qudt/#unit'
```
blushi, line 233: eventually we'll need to update the old projects metadata to use the most up-to-date qudt reference (without the version 1.1) but that can be addressed at a later point
```suggestion
    ]?.['@value'] ||
    metadata?.['http://regen.network/projectSize']?.[
      'http://qudt.org/schema/qudt/unit'
```

web-registry/src/lib/rdf.ts
```diff
@@ -58,8 +58,12 @@ export function getProjectPageBaseData(): any {
...
   | 'http://qudt.org/1.1/vocab/unit#HA'
-  | 'http://qudt.org/1.1/vocab/unit#AC';
+  | 'http://qudt.org/1.1/vocab/unit#AC'
+  | 'unit:HA'
+  | 'unit:AC';
 export const qudtUnitMap = {
   'http://qudt.org/1.1/vocab/unit#HA': 'hectares',
   'http://qudt.org/1.1/vocab/unit#AC': 'acres',
+  'unit:HA': 'hectares',
+  'unit:AC': 'acres',
```
blushi, line 68:
```suggestion
  'http://qudt.org/vocab/unit/HA': 'hectares',
  'http://qudt.org/vocab/unit/AC': 'acres',
```

web-registry/src/lib/rdf.ts
```diff
@@ -58,8 +58,12 @@ export function getProjectPageBaseData(): any {
 
 export type qudtUnit =
   | 'http://qudt.org/1.1/vocab/unit#HA'
-  | 'http://qudt.org/1.1/vocab/unit#AC';
+  | 'http://qudt.org/1.1/vocab/unit#AC'
+  | 'unit:HA'
+  | 'unit:AC';
```
blushi, line 63:
```suggestion
  | 'http://qudt.org/vocab/unit/HA'
  | 'http://qudt.org/vocab/unit/AC';
```

mhagel:
@Mergifyio backport redwood

mhagel:
@Mergifyio backport hambach

PR #870, opened 2022-03-21
Victor/847 portfolio basket ledger

victordelval:
## Description

Closes: regen-network/regen-registry#847

The approach used is to extract from the UI component, the logic related to the data of the baskets (request, prepare, state) to custom hooks.

Regarding the hooks, the approach is the following. There are 2 atomic hooks that correspond directly to two ledger api queries:
* query for Baskets (all baskets) => `useQueryBaskets`
* query for Balance (one basket) => `useQueryBalance`

These hooks are used by the `useBasketTokens` business hook that prepares the necessary data for the UI components, specifically for the `<BasketsTable />` component.

Also:
* Moved the formatNumber( ) utility function that was in the table components, to the format utility library.
* Ecocredit data logic moved to a custom hook `useEcocredits`

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

- [ ] reviewed tests


web-registry/src/hooks/useQueryBalance.ts
```diff
@@ -0,0 +1,46 @@
...
+} from '@regen-network/api/lib/generated/cosmos/bank/v1beta1/query';
+
+import { useLedger } from '../ledger';
+
+type FetchBalance = {
+  address: string | undefined;
+  denom: string | undefined;
+};
+
+export default function useQueryBalance(): Function {
```
haveanicedavid, line 15: `Function` isn't the right type here. I believe it should be something like 

```ts
type FetchBalance = (args: Balance) => QueryBalanceResponse // string or whatever the return type of `await queryClient.Balance({ address, denom })` is
``` 
(note: i'm reusing your naming for the returned function)

which would change this line to
```suggestion
export function useQueryBalance(): FetchBalance {
```
or inline
```suggestion
export function useQueryBalance(): (args: Balance) => QueryBalanceResponse { 
```

and give type safety where it's invoked, ie here:
```ts
const fetchBalances = useQueryBalance()
fetchBalances({ address: 'someAddress', denoms: 'someDenom' })
```
you'd get type support with the instantiated `fetchBalances`. Typing as `Function` allows anything to be passed to it without TS complaining

web-registry/src/hooks/useBasketTokens.ts
```diff
@@ -0,0 +1,46 @@
...
+export interface BasketTokens {
+  basket: Basket;
+  balance?: QueryBalanceResponse | undefined;
+  displayDenom: string;
+}
+
+export default function useBasketTokens(
+  address: string | undefined,
+): BasketTokens[] {
+  const baskets = useQueryBaskets();
```
haveanicedavid, line 19: I think it's fine here, but I get a bit concerned about hooks calling other hooks... things can quickly spiral to a chain of side effects that's hard to manage. In general, I prefer hooks do one thing specifically and get external dependencies passed through arguments (ie `useBasketTokens(address, baskets)`)

No need to change this though, I don't think. And I love the use of hooks to DRY up code!

web-registry/src/hooks/useBasketTokens.ts
```diff
@@ -0,0 +1,46 @@
...
+import useQueryBalance from './useQueryBalance';
+
+export interface BasketTokens {
+  basket: Basket;
+  balance?: QueryBalanceResponse | undefined;
+  displayDenom: string;
+}
+
+export default function useBasketTokens(
+  address: string | undefined,
```
haveanicedavid, line 17:
```suggestion
  address?: string,
```

web-registry/src/hooks/useEcocredits.ts
```diff
@@ -0,0 +1,27 @@
+import { useState, useEffect } from 'react';
+
+import { ledgerRESTUri } from '../lib/ledger';
+import { getEcocreditsForAccount } from '../lib/ecocredit';
+import type { BatchInfoWithBalance } from '../types/ledger/ecocredit';
+
+export default function useEcocredits(
+  address: string | undefined,
```
haveanicedavid, line 8:
```suggestion
  address?: string,
```

web-registry/src/hooks/useQueryBaskets.ts
```diff
@@ -0,0 +1,26 @@
...
+  useEffect(() => {
+    if (!api?.queryClient) return;
+
+    const queryClient: QueryClientImpl = new QueryClientImpl(api.queryClient);
+    queryClient
+      .Baskets({})
+      .then(setBaskets)
+      /* eslint-disable */
+      .catch(console.error);
+  }, [api?.queryClient]);
```
haveanicedavid, line 23: I wasn't aware you could use optional chaining in effect dependencies! I expected this to error (i've run into this before and had to extract the optionally-chained value to a variable and use that for the dep array), but it seems to be working for me. Maybe that was a typescript bug that React fixed. Good to know!

haveanicedavid (commented):
A few comments, but overall I like the idea of extracting this logic to hooks!

web-components/src/utils/format.ts
```diff
@@ -49,3 +49,9 @@ export function formatDate(
 ): string {
   return dayjs(date).format(format);
 }
+
+export function formatNumber(num: number | string | undefined): string {
+  if (!num) return '-';
+  if (typeof num === 'string') num = parseFloat(num);
+  return num > 0 ? Math.floor(num).toLocaleString() : '-';
+}
```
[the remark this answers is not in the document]
victordelval: Thanks, I appreciate the thought. The thing is that the change causes me complaints from the TS compiler. I'm going to leave the feature as it is for now (given due dates, etc). We'll be back to this soon.

image

image

victordelval:
I'm refactoring the PortfolioTemplate component, since when removing the data logic it loses a bit of the sense it had as a container.

web-registry/src/hooks/useBasketTokens.ts
```diff
@@ -0,0 +1,46 @@
...
+import useQueryBaskets from './useQueryBaskets';
+import useQueryBalance from './useQueryBalance';
+
+export interface BasketTokens {
+  basket: Basket;
+  balance?: QueryBalanceResponse | undefined;
+  displayDenom: string;
+}
+
+export default function useBasketTokens(
```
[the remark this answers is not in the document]
victordelval: I strongly agree with the module level index. In fact, I was going to do it, but I still have some doubts about the structure of this hooks module... Not so much, or I'm still not sure, with not exporting as default the component or code that corresponds to the file as 1 to 1. Anyway, we will deal with it soon. For now, since the hooks directory is new, I'm going to leave it with default export for each hook, and add the index file to export at the module level. Thanks!!

```
export { default as useBasketDetails } from './useBasketDetails';
export { default as useBasketTokens } from './useBasketTokens';
export { default as useEcocredits } from './useEcocredits';
```

web-registry/src/hooks/useQueryBalance.ts
victordelval: Thank you very much for your comments, now the function returned by the hook is better typed, thanks!

web-registry/src/hooks/useQueryBalance.ts
```diff
@@ -0,0 +1,46 @@
...
+  QueryBalanceResponse,
+  QueryClientImpl,
+} from '@regen-network/api/lib/generated/cosmos/bank/v1beta1/query';
+
+import { useLedger } from '../ledger';
+
+type FetchBalance = {
+  address: string | undefined;
+  denom: string | undefined;
+};
```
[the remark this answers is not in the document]
victordelval: nice, thanks!!

web-registry/src/hooks/useBasketTokens.ts
victordelval: Things can always get messy and difficult to maintain if there is no design behind it ;P Here the idea, at least initially, is to have a layer of primitive queries respecting the RegenAPI and the DTOs it returns, and above that, an abstraction layer of our app's business for handling and preparing the received data. Soon we will begin to review in depth issues of architecture and data management in the application.

Anyway I understand the intention of the dependency injection that you suggest, but not in this case I would say. Or do you think that the ultimate responsibility for preparing the BasketTokens data lies with the UI component that uses the hook?

Thanks, David!!

blushi (changes requested):
Could you update the basket table according to https://www.figma.com/file/x5vjWsddiUBzP2N13AFOPw?node-id=420:17241#167911289?

web-registry/src/hooks/useBasketTokens.ts
```diff
@@ -0,0 +1,44 @@
...
+    async function fetchData(baskets: QueryBasketsResponse): Promise<void> {
+      const _basketTokens = await Promise.all(
+        baskets.baskets.map(async basket => ({
+          basket,
+          balance: await fetchBalance({
+            address,
+            denom: basket.basketDenom,
+          }),
+          // TODO: fetch display denom for basket
+          displayDenom: basket.basketDenom,
```
blushi, line 33: why is this still TODO?
you should use `DenomMetadata` query from cosmos/bank

blushi (changes requested):
We shouldn't show the basket row if the user has no basket tokens on it (can be seen on https://deploy-preview-870--regen-registry.netlify.app/ecocredits/dashboard)

web-registry/src/hooks/useBasketTokens.ts
```diff
@@ -0,0 +1,48 @@
...
+import {
+  QueryBalanceResponse,
+  QueryDenomMetadataResponse,
+} from '@regen-network/api/lib/generated/cosmos/bank/v1beta1/query';
+
+import useQueryBaskets from './useQueryBaskets';
+import useQueryBalance from './useQueryBalance';
+import useQueryDenomMetadata from './useQueryDenomMetadata';
+
+export interface BasketTokens {
```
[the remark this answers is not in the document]
blushi: I think that makes sense to separate those out into different fields anyway

web-registry/src/hooks/useEcocredits.ts
```diff
@@ -0,0 +1,27 @@
+import { useState, useEffect } from 'react';
+
+import { ledgerRESTUri } from '../lib/ledger';
+import { getEcocreditsForAccount } from '../lib/ecocredit';
+import type { BatchInfoWithBalance } from '../types/ledger/ecocredit';
+
+export default function useEcocredits(
```
mhagel, line 7: I like this, but something to think about for naming conventions... React hooks conventionally are called `useWhatever`, but in this case we lose some of the context for what the hook _does_. Maybe something like `useQueryEcocredits` or `useGetEcocredits` would make sense? I see you did this in other hooks.


No need to change here, but just a thought.

mhagel (approved):
Nice work! Some comments, but nothing critical.

web-registry/src/hooks/useEcocredits.ts
victordelval: effectively! 

If you notice, the hooks that I am using with `useQuery...` are atomic and correspond to the api offered by our library (`QueryBaskets` => `useQueryBaskets`) (ultimately what they offer are DTOs, the data returned by the API, without modifying). 

On the other hand, one layer above, are the business hooks of our application, which use the original data to aggregate, transform, etc. to prepare the final data needed by the UI components, therefore, these hooks follow a nomenclature `useBasketTokens(Data)`, `useEcocredits(Data)`, etc., referring only to the required data declaratively.

In any case, all this is a first approach, to organize the logic a bit... undoubtedly something that is open, that will change... and of course we will talk shortly, in the coming weeks.

blushi: ideally, all these hooks could be code generated somehow, and that could be something that some @regen-network/api-react-hooks (just an example name) library could provide, some food for thoughts...

web-registry/src/components/organisms/BasketsTable.tsx
```diff
@@ -33,26 +42,32 @@ export const BasketsTable: React.FC<{
...
             </Grid>
-            <Grid item sx={{ ml: 3.5, fontWeight: 700 }} alignSelf="center">
+            <Grid
+              item
+              sx={{ ml: 3.5, pb: 2, fontWeight: 700 }}
+              alignSelf="center"
+            >
               <Box sx={{ fontSize: theme => theme.spacing(4.5) }}>
-                {row.name}
+                {row.basket.name}
```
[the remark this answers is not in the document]
victordelval: thanks for the comments Mark! I'm resolving these conversations about undefined, though I'm not making any changes to this PR, because I'm incorporating them into the next PR, which is based on this one and adds some refactoring, etc.

web-registry/src/hooks/useEcocredits.ts
victordelval: I'm already seeing how they are repeated in the next PR. We'll talk about it next week. Super!

blushi:
@Mergifyio backport redwood

@Mergifyio backport hambach

blushi:
@Mergifyio backport hambach

PR #892, opened 2022-04-07
David/856 rnct graphic

haveanicedavid:
## Description

- [ ] https://github.com/regen-network/regen-sanity/pull/15 should be merged first

Closes: regen-network/regen-registry#856

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

- [ ] reviewed tests


web-registry/sanity-codegen.yml
```diff
@@ -1,5 +1,5 @@
 overwrite: true
-schema: 'https://jm12rn9t.api.sanity.io/v1/graphql/staging/slugs'
+schema: 'https://jm12rn9t.api.sanity.io/v1/graphql/staging/default'
```
haveanicedavid, line 2: @blushi I want to make sure you see this - I thought because our sanity setup is deploying staging to `/default` that's what we should use, but think I recall you setting it to `/slugs` for a reason - is this okay?

web-registry/src/components/atoms/SanityButton.tsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
```
haveanicedavid, line 1: I built this component to help with one of the things i've noticed with our sanity setup. We have a `onBtnClick` helper which properly manages modal behavior (though it requires a callback, and sometimes buttons are just for links) and `target` - but because that function has no access to `react-router`'s `navigate` function, internal links won't take advantage of the SPA nature of the app (the whole page will re-load). 

I thought about modifying or creating a new helper function to accept `navigate` as an optional param, but think this component makes more sense overall and should simplify some of the boilerplate for adding buttons from sanity. 

The downside is that this won't work in `web-www`. I tried to make something more flexible (which accepted a `navigate` function through props instead of invoking the `useNavigate` hook so we could pass Gatsby's [navigate](https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-link/#how-to-use-the-navigate-helper-function) and use in `web-www`) but it got a bit tricky with types. 

I *think* I might know how to make something like that work through generics... you'd have a `GenericSanityButton` which would live in `web-components` that accepted `navigate` as a prop and was manually passed the button props as a type variable, and invoke it within both `web-registry` and `web-www`. So in registry you'd have something like (pseudo code):

```tsx
import { useNavigate } from 'react-router-dom'
import GenericSanityButton, { GenericSanityButtonProps } from 'web-components'

type Props = Omit<GenericSanityButtonPops, 'navigate'>

function RegistrySanityButton(props: Props) {
  const navigate = useNavigate()
  return (
    <GenericSanityButton {...props} navigate={navigate} />
  )
}
```

....and also have something similar in `web-www`, but I didn't want to go too much outside of the scope of the story

web-registry/sanity-codegen.yml
blushi: yes, keeping this `slugs` tag was an oversight anyway (I had changed that so that we can work on parallel on different sanity changes without interfering with each other), we should use `default` tag by default

haveanicedavid:
@erikalogie added gradient + changed image to a `png`: https://deploy-preview-892--regen-registry.netlify.app/

erikalogie:
Looks good, thanks!

mhagel (approved):

haveanicedavid:

PR #897, opened 2022-04-13
Victor/860 basket details ledger refactor

victordelval:
## Description

Closes: regen-network/regen-registry#860

Resolves the  task to get data for the basket detail page.

This is a PR draft to resolve the conflicts with the master code around the hooks, I have added the following refactoring iteration, creating a generic hook for the `useQueryClient` ledger queries and the functionalities of the ledger API queries have been extracted as async api functions.

The update has already been done with respect to the master branch.

Some upcoming improvements regarding hooks and the Ledger API:
* Improve the DX of the useQueryClient hook (hiding configuration details and/or currying)
* Unify ledger functionalities in its own module (context, types, async functions corresponding to the api, function compositions, hooks, validations, etc)
---

Update - April 25th

A version is finally being reached in which there is a generic hook per ledger module (useBankQuery, useBasketQuery, etc).

```
const { data: basket, loading, error } = useBasketQuery<QueryBasketResponse>({
    queryName: 'basket',
    params: { basketDenom },
});
```

The Ecocredits case is still pending.

Parameter inputs for queries are currently fully typed in hooks. For the response data, the type is passed at the same time as the query hook is used (pending improvement, so that the query hook itself encapsulates the response types)

On the other hand, the /lib directory includes the files (bank, basket) with the queries and the corresponding types that define the apis of the modules.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

- [ ] reviewed tests


web-registry/src/lib/ecocredit.ts
```diff
@@ -346,3 +358,133 @@ export const queryEcoClassInfo = async (
     throw new Error(`Error fetching class info: ${err}`);
   }
 };
+
+/**
+ *
+ * REGEN LEDGER API QUERIES
+ *
+ */
```
blushi, line 366: I don't think this is the best place for having those queries
it looks weird to have some `bank` related queries in a file that's called `ecocredit.ts`

web-registry/src/lib/ecocredit.ts
```diff
@@ -346,3 +358,133 @@ export const queryEcoClassInfo = async (
...
+ *  - DenomsMetadata (TODO)
+ *
+ */
+
+// Denom Metadata
+
+export type QueryDenomMetadataProps = {
+  client: BankQueryClient;
+  denom: string;
+};
```
blushi, line 384: I'm not so fond of this pattern that forces you to redefine the request fields separately. This seems like a lot of unnecessary manual work and prone to errors/oversights. We already have those request types generated and available from regen-js: https://github.com/regen-network/regen-js/blob/9e12047790e7779d8a64872528480f0cfd2870ce/packages/api/src/generated/cosmos/bank/v1beta1/query.ts#L1212-L1213

My suggestions:
1. Declare an interface that wraps the `client` that can be extended for defining individual query props:
```
interface BankQueryClientProps {
 client: BankQueryClient;
}
```
2. Use the generated types from regen-js:
```
interface QueryDenomMetadataProps extends BankQueryClientProps {
 request: DeepPartial<QueryDenomMetadataRequest>,
}
```

Another idea would be define take the "extension" approach that is used in cosmjs, see for instance: 
bank extension: https://github.com/cosmos/cosmjs/blob/ae06012a1510ddf48068bbf21374c0bbff3d5bab/packages/stargate/src/modules/bank/queries.ts
how it's being used: https://github.com/cosmos/cosmjs/blob/ae06012a15/packages/stargate/src/stargateclient.ts#L162-L169

web-registry/src/components/organisms/BasketOverview.tsx
```diff
@@ -11,14 +11,105 @@ import Description from 'web-components/lib/components/description';
...
+                  data={curator}
+                  link={getAccountUrl(curator as string)}
+                />
+                <ItemWithLinkList
+                  label="allowed credit classes"
+                  data={allowedCreditClasses}
+                  link={'/credit-classes/'}
+                />
+                <Item
+                  label="min start date"
```
blushi, line 97: this label should be either "min start date" or "start date window"

web-registry/src/hooks/useQueryLedger.ts
```diff
@@ -0,0 +1,106 @@
...
+type OutputQueryLedger<T> = {
+  data: T | undefined;
+  loading: boolean;
+  error: Error | undefined;
+};
+
+export default function useQueryLedger<T>({
+  queryType,
+  queryCallback,
+  queryParams,
```
blushi, line 57: we could just use `type`, `callback` and `params`, using this `query` prefix is a bit redundant.

also this design doesn't refrain the developer from specifying a query type like 'bank' and then trying to call a query callback that is not related to bank

web-registry/src/lib/ecocredit.ts
victordelval: Yes I like this. The proposal brings a lot of change (structure, pattern), in the details there is still room for improvement of course. In the useQueryLedger hook I was able to define the output a bit better with generic types but the input part is also a WIP..

web-registry/src/components/organisms/BasketOverview.tsx
victordelval: yes, it was not yet, I include it

web-registry/src/hooks/useQueryLedger.ts
victordelval: yes, totally... I also thought about changing it at some point. I was also about to encapsulate all those details in the useQueryLedger hook itself so that you only have to specify the query name and query parameters.

web-registry/src/hooks/useQueryLedger.ts
```diff
@@ -0,0 +1,106 @@
+import { useState, useEffect } from 'react';
```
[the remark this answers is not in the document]
victordelval: seems fine to me, thanks

web-registry/src/lib/ecocredit.ts
victordelval: Do you think it's okay if it's something we take into consideration and improve in future iterations/tasks?

web-registry/src/lib/ecocredit.ts
victordelval: Do you think it's okay if it's something we take into consideration and improve in future iterations/tasks?

blushi:
Somehow, there's no credit batch displayed on the eco.uC.rNCT basket page https://deploy-preview-897--regen-registry.netlify.app/baskets/eco.uC.rNCT while there should be some: http://redwood.regen.network:1317/regen/ecocredit/basket/v1/baskets/eco.uC.rNCT/balances so something is wrong there.

image

web-registry/src/lib/ecocredit.ts
blushi: I believe that wouldn't be to much work to move these to another file...

web-registry/src/lib/ecocredit.ts
blushi: I'm not sure, I'll need to think through this a bit more
There are still a lot of // TODO in the `useQueryClient` file with lots of `any` type definition which doesn't seem right to merge as is IMO
But I'd welcome other opinions from @haveanicedavid and @mhagel

victordelval:
> Somehow, there's no credit batch displayed on the eco.uC.rNCT basket page https://deploy-preview-897--regen-registry.netlify.app/baskets/eco.uC.rNCT while there should be some: http://redwood.regen.network:1317/regen/ecocredit/basket/v1/baskets/eco.uC.rNCT/balances so something is wrong there.
> 
> image

after checking for a while today, I've finally updated the branch code with a couple of console.logs for debugging. It looks like the `fetchProjectByBatchDenom` lazy hook query, to retrieve the project name from the credit vintages in the registry DB, is returning null for each batch. Perhaps it is because the corresponding data does not exist (in the credit_vintage or project tables) in the staging registry server DB?

blushi:
> after checking for a while today, I've finally updated the branch code with a couple of console.logs for debugging. It looks like the fetchProjectByBatchDenom lazy hook query, to retrieve the project name from the credit vintages in the registry DB, is returning null for each batch. Perhaps it is because the corresponding data does not exist (in the credit_vintage or project tables) in the staging registry server DB?

Yeah they might not be corresponding `credit_vintage` in the db for each on-chain credit batches so in this case we should just handle it gracefully and not display the project at all.

web-registry/src/hooks/useBasketDetails.ts
```diff
@@ -1,121 +1,235 @@
...
-          (acc: number, obj: BatchInfoWithProject) =>
-            acc + Number(obj.total_amount),
-          0,
+            // eslint-disable-next-line no-console
+            console.log('** batchProject', batchProject);
+
+            return {
+              batchDenom,
+              projectName:
+                batchProject.data!.creditVintageByBatchDenom!
```
blushi, line 150: please use `?` instead of `!` because this can be null or undefined

victordelval: Perfect. Yesterday reviewing the models in graphql I realized that many fields of the project entity are not mandatory... I update the code.

web-registry/src/hooks/useQueryClient.ts
```diff
@@ -0,0 +1,104 @@
...
+// TODO: map clients and ledger queries here, hidding this details from business hooks
+// TODO: Map the clients and queries here, so that the implementation details
+// of the queries are transparent to business hooks. Instead, the hook would be called
+// like this:
+// ie. useQueryClient('basket', { basketDenom })
+// ie. useQueryClient('denomMetadata', { denom: basketDenom })
+
+type QueryClient = BankQueryClient | EcocreditQueryClient | BasketQueryClient;
+
+type LedgerModules = 'bank' | 'ecocredit' | 'basket';
```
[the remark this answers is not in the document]
mhagel: Although I would make it singular
```suggestion
type LedgerModule = 'bank' | 'ecocredit' | 'basket';
```

web-registry/src/hooks/useQueryClient.ts
```diff
@@ -0,0 +1,104 @@
...
+// }: {
+//   client: QueryClient;
+//   params: Ledgerparams;
+// }) => Promise<any>;
+
+// TODO: generic types
+type LedgerQuery = ({ client, params }: any) => Promise<any>;
+
+type InputQueryLedger = {
+  type: LedgerModules;
```
mhagel, line 43: `module` is more expressive here, since it is very relevant in the Cosmos world.
```suggestion
  module: LedgerModule;
```

web-registry/src/lib/ecocredit.ts
mhagel: Yeah, I'd appreciate a separate `bank.ts` file. That way we keep following the cosmos module pattern.

web-registry/src/lib/ecocredit.ts
mhagel: commented above. Agree there are some easy wins here.

web-registry/src/lib/ecocredit.ts
```diff
@@ -346,3 +358,133 @@ export const queryEcoClassInfo = async (
...
+ *  - BasketBalance (TODO)
+ *
+ */
+
+// Basket
+
+export type QueryBasketProps = {
+  client: BasketQueryClient;
+  basketDenom: string;
+  // params: {
```
mhagel, line 441: delete?

web-registry/src/pages/BasketDetails.tsx
```diff
@@ -6,12 +6,12 @@ import { BasketOverview, BasketEcocredits } from '../components/organisms';
 
 const BasketDetails: React.FC = () => {
   const { basketDenom } = useParams<{ basketDenom: string }>();
-  const { dataOverview, dataBasketBatches } = useBasketDetails(basketDenom);
+  const data = useBasketDetails(basketDenom);
```
mhagel, line 9: If the basketDenom doesn't exist, there are endless requests to the server, which is not ideal. Can we short-circuit the request loop somehow if a request fails? Otherwise the other requests keep going if a user goes to any URL other than a correct basket_denom.

I'd be willing to defer this to an error-handling story, if we make sure to document this case.

web-registry/src/hooks/useBasketDetails.ts
```diff
@@ -1,121 +1,220 @@
...
+            return {
+              id: basketClass.info?.classId || '-',
+              name: basketClass.info?.classId || '-',
+            };
+          }),
+        );
+        setBasketClasses(_basketClasses);
       } catch (error) {
-        // TODO Manage error
         console.error(error); // eslint-disable-line no-console
```
mhagel, line 115: could we do some simple error-handling to make sure there aren't endless requests if the ledger can't find something in this file? Like a `[error, setError] = useState(false)`  and then check for `error` in each `useEffect`?

mhagel (changes requested):
This is really close, but it would be nice to short-circuit the request-cycle if the ledger does not return a valid result.

web-registry/src/lib/ecocredit.ts
victordelval: Done! I have moved the functions to their corresponding files and default location `src/lib/bank.ts` and `src/lib/basket.ts`. 

I would suggest taking this further and creating a ledger module that encapsulates all of the ledger-related logic, as I quickly showed the other day at the architecture meeting. Something like this:

`src/ledger/`

image

victordelval:
Related to the suggested improvement of extracting the `bank` and `basket` API functions that were provisionally in the `src/lib/ecocredit.ts` file

I would suggest taking this further and creating a ledger module that encapsulates all of the ledger-related logic, as I quickly showed the other day at the architecture meeting. Something like this:

`src/ledger/`

image

web-registry/src/hooks/useQueryClient.ts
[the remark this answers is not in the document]
victordelval: thanks! well seen!

web-registry/src/hooks/useQueryClient.ts
blushi: agreed

web-registry/src/pages/BasketDetails.tsx
blushi: I believe this should be addressed before we merge this

web-registry/src/hooks/useQueryClient.ts
```diff
@@ -0,0 +1,104 @@
...
+import { QueryClientImpl as BasketQueryClient } from '@regen-network/api/lib/generated/regen/ecocredit/basket/v1/query';
+
+import { useLedger } from '../ledger';
+
+// TODO: map clients and ledger queries here, hidding this details from business hooks
+// TODO: Map the clients and queries here, so that the implementation details
+// of the queries are transparent to business hooks. Instead, the hook would be called
+// like this:
+// ie. useQueryClient('basket', { basketDenom })
+// ie. useQueryClient('denomMetadata', { denom: basketDenom })
```
blushi, line 14: as discussed on slack, this wouldn't work: https://regen-network.slack.com/archives/C01LX9E8QN8/p1649866623584679?thread_ts=1649865324.300319&cid=C01LX9E8QN8 so remove or update this comment

web-registry/src/hooks/useQueryClient.ts
```diff
@@ -0,0 +1,104 @@
...
+
+type QueryClient = BankQueryClient | EcocreditQueryClient | BasketQueryClient;
+
+type LedgerModules = 'bank' | 'ecocredit' | 'basket';
+
+// TODO: generic type
+type Ledgerparams = any;
```
blushi, line 21: let's use camelCase

web-registry/src/hooks/useQueryClient.ts
```diff
@@ -0,0 +1,104 @@
...
+// TODO: generic types
+type LedgerQuery = ({ client, params }: any) => Promise<any>;
+
+type InputQueryLedger = {
+  type: LedgerModules;
+  callback: LedgerQuery;
+  params?: Ledgerparams;
+};
+
+type OutputQueryLedger<T> = {
```
blushi, line 48: maybe we can just get rid of all these "Ledger" prefixes and suffixes in types:
- LedgerQuery -> Query
- OutputQueryLedger -> QueryOutput
- InputQueryLedger -> QueryInput
- LedgerParams -> Params
- LedgerModules -> Module

web-registry/src/lib/ecocredit.ts
blushi: unresolving, this file needs to be cleaned up or you just forgot to push your changes...

web-registry/src/hooks/useQueryClient.ts
blushi: or just get rid of the Ledger prefix as suggested in my comment below

web-registry/src/hooks/useQueryClient.ts
victordelval: yes, totally agree. I have also changed the name of the callback parameter to query, following the same logic

web-registry/src/lib/ecocredit.ts
victordelval: the changes were uploaded, but I hadn't removed the code from the previous version (rest api). Now I have already cleaned the ecocredit files in /types and /lib. do you mean that?

web-registry/src/hooks/useQueryClient.ts
victordelval: that seems fine to me!

web-registry/src/pages/BasketDetails.tsx
victordelval: Fixed up. Now we throw the error from the try/catch of the ledger api query functions. And in the hooks we set the error status and we take it into account in the effects that manage the requests (and it is also exported from the hook in case you want to manage it in the UI)

web-registry/src/hooks/useBasketDetails.ts
victordelval: Done! (See above)

web-registry/src/pages/BasketDetails.tsx
mhagel: Thanks. Tested and looks good now.

mhagel (approved):
Nice work!

blushi (changes requested):

blushi:
image

This account link uses the name: https://redwood.regen.aneka.io/accounts/Regen%20Network%20Development,%20Inc., it should use an address (just hardcode any address for now as we had discussed)

web-registry/src/hooks/useBasketDetails.ts
```diff
@@ -1,121 +1,210 @@
...
-            acc + Number(obj.total_amount),
-          0,
+            return {
+              batchDenom,
+              projectName:
+                batchProject.data?.creditVintageByBatchDenom?.projectByProjectId
+                  ?.handle || '-',
+              projectDisplay:
+                batchProject.data?.creditVintageByBatchDenom?.projectByProjectId
+                  ?.metadata['http://schema.org/name'] || '-',
```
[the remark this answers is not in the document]
victordelval: Done, thanks!

web-registry/src/lib/bank.ts
```diff
@@ -0,0 +1,160 @@
...
+  client,
+  request,
+}: QueryAllBalancesProps): Promise<DeepPartial<QueryAllBalancesResponse>> => {
+  try {
+    return await client.AllBalances({
+      address: request.address,
+    });
+  } catch (err) {
+    throw new Error(
+      'Error in the AllBalances query of the ledger bank module.',
```
blushi, line 95: I think it'd be good to actually use the `err` in `new Error(...)`, otherwise the error msg is not super helpful
ditto for all the similar error handling

victordelval: Seem right. Just concatenating the `err` to the string, right?

web-registry/src/hooks/useBankQuery.ts
```diff
@@ -0,0 +1,110 @@
...
+      response = balance(client, params);
+    }
+
+    if (query === 'denomMetadata') {
+      response = denomMetadata(client, params);
+    }
+
+    if (query === 'denomsMetadata') {
+      response = denomsMetadata(client, params);
+    }
```
blushi, line 88: why not using a switch/case here? and for the default case `setError`?

web-registry/src/lib/bank.ts
```diff
@@ -0,0 +1,162 @@
...
+// All Balances
+
+interface QueryAllBalancesProps extends BankQueryClientProps {
+  request: DeepPartial<QueryAllBalancesRequest>;
+}
+
+export const queryAllBalances = async ({
+  client,
+  request,
+}: QueryAllBalancesProps): Promise<DeepPartial<QueryAllBalancesResponse>> => {
```
blushi, line 88: you don't need the `DeepPartial` in the response type
ditto for all the query functions

web-registry/src/hooks/useBankQuery.ts
victordelval: better thanks!

web-registry/src/lib/bank.ts
victordelval: thanks!

web-registry/src/hooks/useBankQuery.ts
```diff
@@ -67,24 +67,28 @@ export default function useBankQuery<T>({
...
+        response = balance(client, params);
+        break;
+      case 'denomMetadata':
+        response = denomMetadata(client, params);
+        break;
+      case 'denomsMetadata':
+        response = denomsMetadata(client, params);
+        break;
+      default:
+        setError(
```
blushi, line 87: we should also probably set loading back to false in this case

web-registry/src/hooks/useBankQuery.ts
```diff
@@ -0,0 +1,114 @@
...
+          new Error(
+            'You need to provide a valid bank query name (ie. allBalances)',
+          ),
+        );
+    }
+
+    if (response) {
+      response
+        .then(response => setData(response as unknown as T))
```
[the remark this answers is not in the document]
blushi: ditto for your other query hooks

web-registry/src/hooks/useBankQuery.ts
victordelval: yeah, thanks!

web-registry/src/hooks/useBankQuery.ts
```diff
@@ -0,0 +1,114 @@
...
+// TODO - this hook is still missing batch query functionality
+// TODO - this hook is still missing lazy query functionality
+
+type QueryOutput<T> = {
+  data: T | undefined;
+  loading: boolean;
+  error: Error | undefined;
+};
+
+export default function useBankQuery<T>({
```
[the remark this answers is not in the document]
victordelval: this is wonderful!

web-registry/src/hooks/useBankQuery.ts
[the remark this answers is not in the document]
victordelval: thanks a lot!

blushi:
> One question @blushi , do you know how it could be done so that the typing is all hidden in the query hooks, in such a way that when using these hooks from another place (another business hook or a UI component) it is not necessary to inform (import) the type that expected
> 
> ```
> // from some business hook or ui component
> 
> // instead of passing the specific response type that is expected
> // const { data, loading, error } = useBasketQuery<QueryBasketResponse>(...)
> 
> const { data, loading, error } = useBasketQuery({ query: 'basket', params: { basketDenom } });
> ```
> 
> the following happens when the response type is not passed
> 
> query hook response type
> 
> query hook response

I don't see any other way to handle that off the top of my head. I think it's fine to pass in the response type (although I agree it's a bit redundant since we're also passing the query as a string param). I can try to think of some better ways to handle that...

blushi (approved):

victordelval:
@mergifyio backport redwood hambach

PR #936, opened 2022-05-09
Update project forms to use compacted JSON-LD

blushi:
## Description

Closes: https://github.com/regen-network/regen-registry/issues/904

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

- [ ] reviewed tests

### DIFF
--- a/web-registry/src/lib/transform.ts
+++ b/web-registry/src/lib/transform.ts
@@ -117,7 +117,7 @@ export function buildIssuanceModalData(
         creditVintage.endDate,
       ),
       monitoringPeriods,
-      projectName: project.metadata?.['http://schema.org/name'] || '',
+      projectName: project.metadata?.['schema:name'] || '',
       standard: {
         documentId:
           creditClassVersion?.metadata?.['http://regen.network/standard']?.[
--- a/web-registry/src/lib/transform.ts
+++ b/web-registry/src/lib/transform.ts
@@ -200,9 +200,10 @@ const getPartyFromMetadata = (
     type: metadataRole?.['@type'].includes('regen:Organization') // covers Organization or OrganizationDisplay
       ? 'ORGANIZATION' // to provide default image
       : '',
+    image: metadataRole?.['schema:image']?.['@value'],
+    address: metadataRole?.['schema:location']?.place_name || '',
     individual: '',
     role: '',
-    address: '',
   };
 };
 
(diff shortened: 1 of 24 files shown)


web-registry/src/lib/transform.ts
```diff
@@ -200,9 +200,10 @@ const getPartyFromMetadata = (
     type: metadataRole?.['@type'].includes('regen:Organization') // covers Organization or OrganizationDisplay
       ? 'ORGANIZATION' // to provide default image
       : '',
+    image: metadataRole?.['schema:image']?.['@value'],
+    address: metadataRole?.['schema:location']?.place_name || '',
```
blushi, line 204: not directly related but realized this was missing cc/ @mhagel

web-registry/src/lib/transform.ts
```diff
@@ -117,7 +117,7 @@ export function buildIssuanceModalData(
         creditVintage.endDate,
       ),
       monitoringPeriods,
-      projectName: project.metadata?.['http://schema.org/name'] || '',
+      projectName: project.metadata?.['schema:name'] || '',
```
blushi, line 120: not directly related but when the project metadata didn't get updated to use the compacted form in here, fixed it in a few other places as well cc/ @mhagel

victordelval:
Seem right. Nice job!

Just a question that I throw in the air in case it is of interest so that we give it one thought for a next refactor or whatever... 

would it make sense to have a utility or something that parses this type of json-LD structures to avoid using those string notations with semicolons on the keys, to improve DX? 

For example, instead of using a key string like `metadataRole?.['schema:name']` we could use something more JS friendly like `metadataRole?.schemaName` or something like this... 

It would also act as a wrapper to hide the final implementation and avoid having to change in several places if there is an update, etc.

victordelval (approved):

mhagel (approved):
Whew! That's better :)

blushi:
> Seem right. Nice job!
> 
> Just a question that I throw in the air in case it is of interest so that we give it one thought for a next refactor or whatever...
> 
> would it make sense to have a utility or something that parses this type of json-LD structures to avoid using those string notations with semicolons on the keys, to improve DX?
> 
> For example, instead of using a key string like `metadataRole?.['schema:name']` we could use something more JS friendly like `metadataRole?.schemaName` or something like this...
> 
> It would also act as a wrapper to hide the final implementation and avoid having to change in several places if there is an update, etc.

Something to keep in mind.
The only concern that I have is that on input value change, we'll still need to convert the data back to proper JSON-LD for the SHACL validation to work or maintain 2 copies of the data (the one with nicer key names and the JSON-LD one), so I'm not sure if it's worth it.

blushi:

PR #1180, opened 2022-08-31
Feat 1098 show 3 more projects

victordelval:
## Description

Closes: regen-network/regen-registry#1098

* Reuse and refactor some hooks and logic from the Projects page
* Create an generic component ProjectCardsSection
* Add a random selection of projects, prioritizing those with sell orders (and excluding the current project)
* Adjust style margins
* Applies to "Featured Projects" section on the Home page
* Applies to the main section on the general page of Projects

---

### Author Checklist

I have...

#### Setting up backport PRs

    ```
    git merge master
    git merge master
    ```

### How to test

1. Go to https://deploy-preview-1180--regen-registry.netlify.app/projects/C01-001

### Reviewers Checklist

I have...

- [ ] reviewed tests


victordelval:
@erikalogie you can check here, thanks!

https://deploy-preview-1180--regen-registry.netlify.app/projects/C01-001

web-registry/src/components/templates/ProjectDetails/hooks/useOtherProjects.ts
```diff
@@ -1,10 +1,71 @@
-import { useMoreProjectsQuery } from '../../../../generated/graphql';
+import { useMemo } from 'react';
+import { QueryProjectsResponse } from '@regen-network/api/lib/generated/regen/ecocredit/v1/query';
+import _ from 'lodash';
```
haveanicedavid, line 3:
```suggestion
import shuffle from ‘lodash/shuffle`
```

Lodash is notoriously bad at code splitting 
I was going to suggest
```tsx
import { shuffle } from ‘lodash`
```

as it’s what we’re doing in most of the app and I thought it would be properly tree-shaked, but found [this thread](https://www.blazemeter.com/blog/import-lodash-libraries) which made me curious, and turns out it wasn’t working correctly. 

Made a pr to fix: https://github.com/regen-network/regen-web/pull/1182

Also, why are we randomizing these values? just to make sure it displays differently?

web-registry/src/components/templates/ProjectDetails/ProjectDetails.config.ts
```diff
@@ -0,0 +1,3 @@
+export const IMAGE_STORAGE_BASE_URL =
+  process.env.REACT_APP_IMAGE_STORAGE_BASE_URL;
+export const API_URI = process.env.REACT_APP_API_URI;
```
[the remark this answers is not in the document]
victordelval: It's actually a convention that we've recently started adopting in folder-driven components, to make large components more readable we're outputting the constants or configuration parameters to a Component.config.ts file.

I take the opportunity to apply it to other files of the ProjectDetails component.

That said, I think it would be good to centralize these global constants at the application level, so that the environment variables are read at a single point in the app, or other general configurations are defined! We'll talk about it soon

web-registry/src/components/templates/ProjectDetails/ProjectDetails.tsx
```diff
@@ -211,9 +213,47 @@ function ProjectDetails(): JSX.Element {
         />
       )}
 
-      {otherProjects && otherProjects.length > 0 && (
-        <div className="topo-background-alternate">
-          <MoreProjectsSection projects={otherProjects} />
+      {otherProjects && (
+        <div id="projects">
```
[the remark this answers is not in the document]
mhagel: yes, `topo-background-alternate` is important to keep to alternate the background on each section

web-registry/src/components/templates/ProjectDetails/hooks/useOtherProjects.ts
victordelval: Nice, thanks!

yes, exactly, to display a random subset

web-registry/src/components/templates/ProjectDetails/ProjectDetails.tsx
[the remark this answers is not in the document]
victordelval: well seen, I have to make a couple of improvements here..

victordelval:
Hey @flagrede! I've refactored it a bit and it affects the Home and the general Projects page (extracting some of the logic and section component). You will tell me what you think. have a good weekend!

erikalogie:
This looks great!

web-registry/src/pages/Home/Home.styles.tsx
```diff
@@ -10,15 +10,15 @@ export const useHomeStyles = makeStyles(theme => ({
...
-  },
+  // projectCards: {
+  //   [theme.breakpoints.down('md')]: {
+  //     width: '100vw',
+  //     marginLeft: theme.spacing(-8),
+  //   },
+  //   [theme.breakpoints.down('sm')]: {
+  //     marginLeft: theme.spacing(-4),
+  //   },
+  // },
```
flagrede, line 21: Should we remove these lines ?

web-registry/src/components/templates/ProjectDetails/hooks/useMoreProjects.ts
```diff
@@ -0,0 +1,56 @@
...
+import { ProjectWithOrderData } from 'pages/Projects/Projects.types';
+import { useProjectsWithOrders } from 'hooks/projects/useProjectsWithOrders';
+
+const PROJECTS_LIMIT = 3;
+
+export default function useMoreProjects(
+  projectId: string,
+): ProjectWithOrderData[] {
+  // get normalized projects with sell order data
+  const { projectsWithOrderData } = useProjectsWithOrders({});
```
flagrede, line 13: `useProjectsWithOrders` is used without a limit here, which means that it will make an individual query to get the metadata for each project while we only need 3 projects.
The logic that picks the 3 random projects should be placed before we do the metadata fetch request to avoid doing unnecessary requests. By doing so we would drastically reduce the number of queries done by the hook.

flagrede (changes requested):
Nice job on the refactoring, it looks very clean 👏
I just have one request change to reduce the number of queries we make to get the 3 random projects.

web-registry/src/components/templates/ProjectDetails/hooks/useMoreProjects.ts
flagrede: This could be a new param to give for the hook:
`useProjectsWithOrders({limit: 3, shuffle: true})`
or it could something more generic to fit more use cases:
`useProjectsWithOrders({limit: 3, sortFunction: () => ...})`
The idea would be to pass a sort function that puts you 3 random projects on top of the array, and then the limit would slice only those 3 projects and it would make only 3 metadata requests.

web-registry/src/pages/Home/Home.styles.tsx
victordelval: thanks!

web-registry/src/components/templates/ProjectDetails/hooks/useMoreProjects.ts
victordelval: I have tried to improve a bit this point, although not so generic. I hope it works for now. Now we make the fair and necessary requests of the projects that we want to show. On the other hand it seems that all these requests were being done twice due to some rerender or something..

web-registry/src/hooks/projects/useProjectsWithOrders.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  const numProjects = limit - projectsWithOrders.length;
+  const additionalProjects = projects.filter(
+    p1 => !projectsWithOrders.some(p2 => p1.id === p2.id),
+  );
+  return [...projectsWithOrders, ...additionalProjects.slice(0, numProjects)];
+};
+
+const isOtherProject = (project: ProjectInfo, projectId: string): boolean =>
+  project.id !== projectId;
```
flagrede, line 41: All these utils functions should probably be in a `useProjectsWithOrders.utils.ts` file to make it more readable.

web-registry/src/hooks/projects/useProjectsWithOrders.ts
```diff
@@ -0,0 +1,101 @@
...
+  projects: ProjectInfo[],
+  sellOrders: SellOrderInfo[],
+  limit?: number,
+): ProjectInfo[] => {
+  const projectsWithOrders = projects.filter(projs =>
+    hasSellOrders(projs, sellOrders),
+  );
+  if (!limit || projectsWithOrders.length >= limit) return projectsWithOrders;
+
+  const numProjects = limit - projectsWithOrders.length;
```
flagrede, line 33: I think this variable will be a negative number most of the time, is this wanted?

web-registry/src/hooks/projects/useProjectsWithOrders.ts
```diff
@@ -0,0 +1,101 @@
...
+  useProjectsSellOrders,
+} from 'pages/Projects/hooks/useProjectsSellOrders';
+import useEcocreditQuery from 'hooks/useEcocreditQuery';
+import { useQuerySellOrders } from 'hooks/useQuerySellOrders';
+
+const hasMetadata = (project: ProjectInfo): boolean => !!project.metadata;
+
+const hasSellOrders = (
+  project: ProjectInfo,
+  sellOrders: SellOrderInfo[],
```
flagrede, line 19: Could all these utils functions use an argument object param `{project, sellOrders}` instead of named parameters? It will be easier to add/remove parameters in the future this way.

web-registry/src/hooks/projects/useProjectsWithOrders.ts
```diff
@@ -0,0 +1,101 @@
...
+  const sellOrders = sellOrdersResponse?.sellOrders;
+
+  const { projectsWithOrderData, loading: loadingWithOrders } =
+    useProjectsSellOrders({
+      projects: selectProjects(projects, sellOrders),
+      sellOrders,
+      limit,
+    });
+
+  function selectProjects(
```
flagrede, line 76: This function should also probably be in `useProjectsWithOrders.utils.ts`. Even if it adds 4 arguments to it, it's best not to have it redefined in each render.

web-registry/src/hooks/projects/useProjectsWithOrders.ts
```diff
@@ -0,0 +1,101 @@
...
+    _projects: ProjectInfo[] | undefined,
+    sellOrders: SellOrderInfo[] | undefined,
+  ): ProjectInfo[] | undefined {
+    if (!_projects || !sellOrders) return;
+
+    if (projectId)
+      _projects = _projects.filter(proj => isOtherProject(proj, projectId));
```
flagrede, line 83: Careful here, it's an assignment to a function parameter that could lead to weird behavior. Would be best to work function scope variable instead.

web-registry/src/hooks/projects/useProjectsWithOrders.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    if (random)
+      return prioritizeWithSellOrders(
```
flagrede, line 86: I don't think calling this function is needed anyways. Once the projects have been shuffled + filtered on metadata, they will be sorted again on sell orders here: https://github.com/regen-network/regen-web/blob/master/web-registry/src/pages/Projects/hooks/useProjectsSellOrders.tsx#L59

Sorting after shuffling should give you the order you want I think. Try to remove it, should give you the same result.

flagrede (commented):
Nice job!
Just a few more minor suggestions/questions and should be good to go.

web-registry/src/hooks/projects/useProjectsWithOrders.ts
victordelval: mmm... actually I think not, because it would have already returned from the function, in the previous check...

In any case, I have significantly changed the function, because I have seen a case that was not entirely correct.

flagrede: You're right, I was confused when reading the code but this will not happen indeed.

web-registry/src/hooks/projects/useProjectsWithOrders.ts
victordelval: mmm, I don't think it's necessary, really. These functions are small utilities, they shouldn't do anything else I think.

victordelval: mm, applying to the main selection function :+1:

web-registry/src/hooks/projects/useProjectsWithOrders.ts
victordelval: hmm, but the thing is we don't want to sort after shuffling. We want 3 random projects, prioritizing those with sell orders... (at the time of implementing this only 2 projects with sell orders in redwood)

flagrede: Yes, that was my point. If you shuffle an array and then sort it on a field, it will still be like picking random projects but with having the one with the field sorted first.
image
When you're writing a sorting function you're not necessarily changing all the element's order.
In this case, we can just put elements with sell order first and leave the rest in the same place in the array.

This is what this sort function is doing: https://github.com/regen-network/regen-web/blob/a1e5bb31afd475adebe8f4a4ab7026a58422a6b2/web-registry/src/pages/Projects/hooks/useProjectsSellOrders.utils.tsx#L74

web-registry/src/hooks/projects/useProjectsWithOrders.ts
flagrede: Since these functions won't probably be reused elsewhere it might be fine. But in general, I'm advocating for using object param by default. Recently I had to do some refactoring on a function without an object parameter and I had to change all function calls because the signature changed even if it was with an optional param, so let's try to avoid that if we can.

web-registry/src/hooks/projects/useProjectsWithOrders.ts
flagrede: The only thing that might be needed on top of shuffling the projects is removing the current one I think.

victordelval: ok, I was not aware that sortProjectsBySellOrdersAvailability ordering was recently introduced in the final step in getProjectDisplayData

web-registry/src/hooks/projects/useProjectsWithOrders.ts
victordelval: Ok, I agree

web-registry/src/hooks/projects/useProjectsWithOrders.ts
victordelval: @flagrede Done! 
I've also fixed the `sortProjectsBySellOrdersAvailability` function, as it was putting projects with sell orders at the end instead of at the beginning. I have also improved the performance by changing `filter` to `some` since it is enough for us to know that a project has a sell order, right?

flagrede: Sounds good, weird that it was reversed but I trust you on this.

mhagel (approved):

victordelval:

PR #1272, opened 2022-09-21
fix: orders with decimals quantity

flagrede:
## Description

Closes: regen-network/regen-registry#1231

image

Before the computing of the total available credits of a project sometimes end up with `NaN` when a project had sell orders with fractional quantity (ie: 0.01).
We're now parsing all order quantities as if they could have decimals.

Note: Ensure if having orders with fractional quantity is legit or if it should be prevented earlier on, will clarify that.

---

### Author Checklist

I have...

#### Setting up backport PRs

    ```
    git merge master
    git merge master
    ```

### How to test

1. Go to https://deploy-preview-1272--regen-registry.netlify.app/
2. Check the credits available on the first project

### Reviewers Checklist

I have...

- [ ] reviewed tests

### DIFF
--- a/web-registry/src/utils/number/format.ts
+++ b/web-registry/src/utils/number/format.ts
@@ -1,12 +1,12 @@
-type RoundFloatNumberParams = {
-  number: number;
+type RoundFloatNumberParamsOptions = {
   decimals?: number;
 };
 
-export const roundFloatNumber = ({
-  number,
-  decimals = 2,
-}: RoundFloatNumberParams): number => {
+export const roundFloatNumber = (
+  number: number,
+  options?: RoundFloatNumberParamsOptions,
+): number => {
+  const decimals = options?.decimals ?? 2;
   const base = Math.pow(10, decimals);
   return Math.round(number * base) / base;
 };
--- a/web-registry/src/utils/number/format.ts
+++ b/web-registry/src/utils/number/format.ts
@@ -1,0 +1,12 @@
+type RoundFloatNumberParams = {
+  number: number;
+  decimals?: number;
+};
+
+export const roundFloatNumber = ({
+  number,
+  decimals = 2,
+}: RoundFloatNumberParams): number => {
+  const base = Math.pow(10, decimals);
+  return Math.round(number * base) / base;
+};
--- a/web-registry/src/pages/Projects/hooks/useProjectsSellOrders.utils.tsx
+++ b/web-registry/src/pages/Projects/hooks/useProjectsSellOrders.utils.tsx
@@ -1,6 +1,6 @@
 import { SellOrderInfo } from '@regen-network/api/lib/generated/regen/ecocredit/marketplace/v1/query';
 import { ProjectInfo } from '@regen-network/api/lib/generated/regen/ecocredit/v1/query';
-import { roundFloatNumber } from 'utils/number/format';
+import { roundFloatNumber } from 'utils/number/format/format';
 
 import { PurchaseInfo } from 'web-components/lib/components/cards/ProjectCard/ProjectCard.types';
 import { formatNumber } from 'web-components/lib/utils/format';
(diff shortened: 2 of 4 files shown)


flagrede:
@erikalogie the available credits are now displayed for all projects: https://deploy-preview-1272--regen-registry.netlify.app/
The problem was that some orders for one project had a fractional quantity. Not sure if this is legit or if we should prevent that. In any case, those projects now display their quantity with decimals up to 2 numbers. Let me know if we should none instead.

erikalogie:
Yes I think that we shouldn't allow fractional quantities of credits. Did not realize that was happening! I guess that should not be allowed as a credit amount in the sell orders popup, the buy popup, the retire, or the send popup. 

Can we simply disable decimal points in those fields? And then add an error message "only whole numbers allowed" if they try to type a decimal point? Similar to this but maybe we can prevent the input of the decimal point altogether? https://www.figma.com/file/seUiQM2SPt78KJiDL0Ea8m/Basic-Ecocredits-Ledger-View-%2F-Marketplace?node-id=4869%3A107869

erikalogie:
@S4mmyb is there a use case for buying fractions of credits?

flagrede:
@erikalogie if we does want to prevent order with decimal point quantity, I'll change the fix in this PR to always round quantity if needed. We can address the new form rule on that in other issue. Is that fine for you ?

web-registry/src/utils/number/format.ts
```diff
@@ -0,0 +1,12 @@
+type RoundFloatNumberParams = {
+  number: number;
+  decimals?: number;
+};
+
+export const roundFloatNumber = ({
+  number,
+  decimals = 2,
+}: RoundFloatNumberParams): number => {
```
haveanicedavid, line 9: Can't remember where I mentioned something similar in another story, but minor nit: I'd prefer having `number` as the first argument and `options` as the second
```tsx
export const roundFloatNumber = (
  num: number,
  options?: { decimals?: number },
): number => {
  const decimals = options?.decimals ?? 2
  const base = Math.pow(10, decimals)
  return Math.round(num * base) / base
}
```

it's a pretty simple helper function, so having to explicitly pass `roundFloatNumber({ number: 1.234 })`  is a bit tedious compared to `roundFloatNumber(1.234)`

haveanicedavid (approved):

web-registry/src/utils/number/format.ts
flagrede: @haveanicedavid  Fair point but I have a counter-argument to that, function without named params tends to be harder to extend. I had a case several times recently where I had to change the signature of functions without named params and I had to update the signature everywhere, which was a bit painful. So even if it seems shorter to write `roundFloatNumber(1.234)` now I would rather have `roundFloatNumber({ number: 1.234 })` to make it more easily extendable in the future.

flagrede: Well after second thinking I think for this case the option solution might be good enough. In anycase it would be probably best to use named parameter by default in favor of extendabilty and use that solution only when it makes sense.

mhagel (approved):

web-registry/src/utils/number/format.ts
wgwz: My take is that in this case, the functions purpose is fairly single-minded, and since it doesn't seem like a large or complex set of dependencies in the function signature, it makes sense to me to have a simple call signature, with options, i.e. `roundFloatNumber(1.234, {decimals: 10})`

web-registry/src/utils/number/format.ts
```diff
@@ -0,0 +1,12 @@
+type RoundFloatNumberParamsOptions = {
+  decimals?: number;
+};
+
+export const roundFloatNumber = (
```
wgwz, line 5: would you be able to add a few unit tests for this @flagrede? i know it's a simple function so i don't consider it super necessary, but for the purpose of understanding what it's doing, seeing some example inputs and outputs would be helpful for me. and perhaps storing those examples as unit tests is nice, rather than just a doc string or something.

wgwz (approved):

web-registry/src/utils/number/format.ts
flagrede: Added here: https://github.com/regen-network/regen-web/pull/1272/commits/0d4d7a9e32acc64d5816d0ec2cb9dd36cac14dd0

flagrede:

PR #1261, opened 2022-09-20
feat: add project name and credit class name to tables

flagrede:
## Description

Closes: regen-network/regen-registry#1072 regen-network/regen-registry#1191 regen-network/regen-registry#1225

- add the project name and credit class name columns to all data tables.
- update hooks and API calls to project name and credit class name.
- add common style for an ellipsis column with approx ~10 char.acters
- prevent unnecessary requests from `useBasketTokens` hook.
- filter credit class column on some pages

---

### Author Checklist

_All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues._

I have...

- [x] provided a link to the relevant issue or specification
- [x] provided instructions on how to test
- [x] reviewed "Files changed" and left comments if necessary
- [ ] confirmed all CI checks have passed
- [ ] once the PR is closed, set up backport PRs for `redwood` and `hambach` (see below)

#### Setting up backport PRs

After merging your PR to `master`, set up backports by doing the following:

1. If your branch does not have merge commits, add the following comment to
   your PR, `@Mergifyio backport redwood hambach`.

2. If your branch does have merge commits:

    a. Pull latest `master`, `hambach` and `redwood` branches

    b. Create new branches for backports and merge `master` (replace `<PR#>` with your PR #)
    ```
    git checkout -b hambach-backport-<PR#> hambach
    git merge master
    git push origin hambach-backport-<PR#>
    git checkout -b redwood-backport-<PR#> redwood
    git merge master
    git push origin redwood-backport-<PR#>`
    ```

    c. Open new PRs in regen-web targeting `hambach` and `redwood`, respectively.
  

### How to test

1. Go to https://deploy-preview-1261--regen-registry.netlify.app/ecocredits/dashboard
2. check new project name + credit class name

### Reviewers Checklist

_All items are required. Please add a note if the item is not applicable and please **add
your handle next to the items reviewed if you only reviewed selected items**._

I have...

- [ ] confirmed all author checklist items have been addressed
- [ ] reviewed code correctness and readability
- [ ] verified React components follow DRY principles
- [ ] reviewed documentation is accurate
- [ ] reviewed tests
- [ ] manually tested (if applicable)

PR #1279, opened 2022-09-22
perf: replace issuer query

flagrede:
## Description

Currently, on redwood, we're doing approximately ~200 requests just to know if a user is an issuer. This is because a lot of credit class has been created recently (we now have about 47).

Capture d’écran 2022-09-22 à 17 56 09

On production, it will less significant since we'll probably have around 4.

I was wondering if instead of using the expensive `useQueryIfIssuer` hook we could just check if the user is an admin on any project?

---

### Author Checklist

I have...

#### Setting up backport PRs

    ```
    git merge master
    git merge master
    ```

### How to test

1.

### Reviewers Checklist

I have...

- [ ] reviewed tests

### DIFF
--- a/web-registry/src/pages/Dashboard/Dashboard.tsx
+++ b/web-registry/src/pages/Dashboard/Dashboard.tsx
@@ -18,7 +18,6 @@ import { useWallet } from 'lib/wallet';
 import { useEcocreditQuery } from 'hooks';
 import { useQueryIfCreditClassAdmin } from 'hooks/useQueryIfCreditClassAdmin';
 import { useQueryIfCreditClassCreator } from 'hooks/useQueryIfCreditClassCreator';
-import { useQueryIfIssuer } from 'hooks/useQueryIfIssuer';
 import { useQueryIfProjectAdmin } from 'hooks/useQueryIfProjectAdmin';
 
 import MyCreditBatches from './MyCreditBatches';
--- a/web-registry/src/pages/Dashboard/Dashboard.tsx
+++ b/web-registry/src/pages/Dashboard/Dashboard.tsx
@@ -1,6 +1,7 @@
 import React, { Suspense } from 'react';
 import { Box, SxProps } from '@mui/material';
 import { useTheme } from '@mui/styles';
+import { QueryProjectsByAdminResponse } from '@regen-network/api/lib/generated/regen/ecocredit/v1/query';
 
 import { Center, Flex } from 'web-components/lib/components/box';
 import { CreditBatchIcon } from 'web-components/lib/components/icons/CreditBatchIcon';
@@ -14,6 +15,7 @@ import { IconTabs } from 'web-components/lib/components/tabs/IconTabs';
 
 import { useWallet } from 'lib/wallet';
 
+import { useEcocreditQuery } from 'hooks';
 import { useQueryIfCreditClassAdmin } from 'hooks/useQueryIfCreditClassAdmin';
 import { useQueryIfCreditClassCreator } from 'hooks/useQueryIfCreditClassCreator';
 import { useQueryIfIssuer } from 'hooks/useQueryIfIssuer';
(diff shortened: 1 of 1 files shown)


web-registry/src/pages/Dashboard/Dashboard.tsx
```diff
@@ -43,11 +44,16 @@ const sxs = {
...
   const isCreditClassAdmin = useQueryIfCreditClassAdmin();
   const isProjectAdmin = useQueryIfProjectAdmin();
-  const walletContext = useWallet();
+  const { wallet } = useWallet();
+  const { data: projectsResponse } =
+    useEcocreditQuery<QueryProjectsByAdminResponse>({
+      query: 'projectsByAdmin',
+      params: { admin: wallet?.address },
+    });
+  const isIssuer = (projectsResponse?.projects.length ?? 0) > 0;
```
wgwz, line 56: could we just put this logic into the `useQueryIfIssuer` hook? i added this hook, so it's definitely done with a bit of naivety.

flagrede:
> @flagrede @wgwz @mhagel @haveanicedavid @erikalogie
> 
> > I was wondering if instead of using the expensive useQueryIfIssuer [hook](https://github.com/regen-network/regen-web/blob/master/web-registry/src/hooks/useQueryProjectsByIssuer.ts) we could just check if the user is an admin on any project (which would be 1 request)? Because apparently, we have a hook useQueryProjectsByIssuer that is calling projectsByAdmin.
> 
> hmm, I think the point is that an issuer does not have to end up being the admin of a project... The issuer is the one who can create projects (associated to the class that the issuer can issue), and by default it is the admin, but normally the role is delegated to an actor closer to the project (the land manager or the project developer)
> 
> right?

So, unfortunately, we might need a `CreditClassByIssuer` call to go around this issue, I guess.

web-registry/src/pages/Dashboard/Dashboard.tsx
flagrede: Sure if using `projectsByAdmin` is sufficient as a replacement. I don't think `useQueryIfIssuer` implementation is naive, it's just that there's no other way to do it with the current ledger version unfortunatly.

wgwz:
@flagrede @victordelval i just wanted to link to this documentation for the ledger, in case it is helpful:

- https://docs.regen.network/modules/ecocredit/01_concepts.html#credit-class-admin
- https://docs.regen.network/modules/ecocredit/01_concepts.html#credit-class-issuers

> The credit class issuers are the addresses with the authority to issue credit batches under the given credit class. The list of credit class issuers is defined at the time the credit class is created and only the admin can update the list after the credit class is created.

so indeed, @victordelval , what you say is right.

> Currently, on redwood, we're doing approximately ~200 requests just to know if a user is an issuer. This is because a lot of credit class has been created recently (we now have about 47).

Shouldn't this be making 47 requests then? Because the way I'm seeing it, it should be 1 request per credit class. Right? If that's the case and we're making some extra requests somehow, could we use a similar approach to what you did here @flagrede ? https://github.com/regen-network/regen-web/pull/1261/commits/594fcdb4a335340a980b2c540b6e9ef3c2e061cf

> So, unfortunately, we might need a CreditClassByIssuer call to go around this issue, I guess. 

@flagrede is this a function that already exists? or it would need to be created?

flagrede:
@wgwz Thanks for the links.
Actually, for each ledger request, there's an additional `status` request made to get the latest block info.
So 47 requests made by these hooks would actually trigger twice the number of HTTP requests.
That being said you're right there's a rerender problem on top of that that triggers the hook several times.

@wgwz @mhagel @victordelval @haveanicedavid 
I updated this PR to fix the over-fetching issue and revert the other change, it's ready to be reviewed again.

@wgwz regarding `CreditClassByIssuer ` this would be a new method on the ledger side that could solve this performance issue completely.

PR #1514, opened 2022-10-24
feat: Keplr mobile support

flagrede:
## Description

Closes: regen-network/regen-registry#786

- handle the connection to Keplr mobile wallet through WalletConnect (both QRCode and In-app link)
- Refactor `Wallet` architecture to support multiple wallets
- Add new UI component `WalletModal` to make wallet selection
- Refactor `Header` component to display `WalletButton` on mobile
- Handle wallet refresh on account change

---

### Author Checklist

I have...

#### Setting up backport PRs

   your PR:

   ```
   git merge dev
   git merge dev
   ```

### How to test

1.

### Reviewers Checklist

I have...

- [ ] reviewed tests


web-components/src/components/block-content/BlockContent.tsx
```diff
@@ -47,6 +47,9 @@ const CustomBlockContent: React.FC<{
         <UnderlineTooltip {...props} title={tooltipText} />
       ),
     },
+    types: {
+      undefined: () => <></>,
```
flagrede, line 51: Fix a console warning

web-components/src/components/header/Header.styles.ts
```diff
@@ -0,0 +1,129 @@
+import { DefaultTheme as Theme, makeStyles } from '@mui/styles';
+
+interface StyleProps {
+  color: string;
+  borderBottom?: boolean;
+  fullWidth: boolean;
+}
+
+export const useHeaderStyles = makeStyles<Theme, StyleProps>((theme: Theme) => {
```
flagrede, line 9: Extract existing Header styles for readability.

flagrede:
@erikalogie You can try mobile support here: https://deploy-preview-1514--regen-registry.netlify.app/

Use cases to test:
- connect on desktop
- connect on mobile with QRCode
- connect on mobile with direct link to app

Note: I wasn't able to test transactions so far because Keplr mobile doesn't support testnets, so we might need to setup a wallet to test on production directly.

erikalogie:
Nice job with this! And good job improvising some of the missing UI screens.

Are you using a wallet connect integration here? I don't ever see their standard UI components as in the mocks.

Connecting on desktop via the mobile qr code isn't working. I get a "this site can't be reached" error. I do have keplr wallet installed on my mobile device and the direct mobile link worked for me, though I can't sign a transaction to create the connection as you already mentioned. 

Sometimes I am getting the qr code screen on mobile, which doesn't really make sense. The user should never get the QR code on a mobile device. Perhaps this is happening because I can't actually connect to keplr wallet on mobile via signing a transaction. I was able to open the keplr mobile wallet the first time using the wallet connect button, but now I seem to always get the qr code when testing on my mobile device.

Can you make the button just say "connect" on mobile so it isn't taking up so much space? Alternatively we can just have both mobile and desktop say "connect".

I think it would be nice if the QR code screen looked more like this on desktop: https://www.figma.com/file/2ORMWGHMsB1rzGMrRIAtEV/Keplr-Log-In?node-id=731%3A33214

flagrede:
@erikalogie thanks for the feedback.

Yes, we're using wallet connect but only for the logic part, we don't have to use their components here.

Sorry for the issues I should have made it more clear in the meeting. I realized yesterday while showing the PR in the meeting that the deployment is not going to work for testing because it runs on redwood. Keplr mobile does not support custom chains like testnets. In order for you to test we would to run a deployment with production environment variables (not sure how to do that yet).

Regarding the QR code issue, I think you also had that because of the problem mentioned above.
I tried to mirror Osmosis UX regarding when we should display QRCode or not but yes you should not see it on a mobile device.

Will update the wording, and the QR screen.
Alternatively for the wording, we could keep "connect wallet" on mobile but with a smaller fontsize, wdyt?

flagrede:
@erikalogie I've setup a sandbox to deploy with production configuration: https://alike-spring.surge.sh/
You should be able able to test the mobile connection now.
I've also handles your feedback on this deploy.

web-registry/src/components/organisms/WalletButton/WalletButton.tsx
```diff
@@ -0,0 +1,79 @@
+import React, { useCallback, useMemo, useState } from 'react';
+
+import ErrorBanner from 'web-components/lib/components/banner/ErrorBanner';
+import OutlinedButton from 'web-components/lib/components/buttons/OutlinedButton';
+import WalletModal from 'web-components/lib/components/modal/wallet-modal';
+import { WalletModalState } from 'web-components/lib/components/modal/wallet-modal/WalletModal.types';
+
+import Keplr from '../../../assets/keplr.png';
```
wgwz, line 8: should we use absolute imports here?

web-components/src/components/modal/wallet-modal/components/WalletModal.Mobile.config.tsx
```diff
@@ -0,0 +1,43 @@
...
+  uri?: string;
+};
+
+export const getWalletModalMobileTabs = ({
+  mobileConnectUrl,
+  uri,
+}: getWalletModalMobileTabsParams): IconTabProps[] =>
+  [
+    mobileConnectUrl
+      ? {
```
wgwz, line 19: could we switch this to something other than a ternary operator? i find ternary operators that involve a bigger chunk of code like this hard to read.

web-registry/package.json
```diff
@@ -62,9 +68,8 @@
     "lint-fix": "eslint ./src --fix -c .eslintrc.js --ext .ts,.tsx",
     "format": "prettier --write --loglevel warn './src/**/*.{ts,tsx,json,md,css}'",
     "format-and-fix": "yarn format && yarn lint-fix",
-    "start": "react-scripts start",
-    "build": "react-scripts build",
-    "test": "react-scripts test",
+    "start": "react-app-rewired start",
```
wgwz, line 71: curious to know what this change is for?

web-components/src/components/modal/wallet-modal/components/WalletModal.Mobile.config.tsx
wgwz, line 19: similar thing here to another comment i added. i'd be grateful to have this without the ternary, if possible/you don't mind.

web-registry/src/lib/wallet/wallet.utils.ts
```diff
@@ -0,0 +1,85 @@
...
+  const walletConnect = new (await import('@walletconnect/client')).default({
+    bridge: 'https://bridge.walletconnect.org',
+    signingMethods: [
+      'keplr_enable_wallet_connect_v1',
+      'keplr_sign_amino_wallet_connect_v1',
+    ],
+    qrcodeModal: {
+      open: (uri: string, callback: () => void) => {
+        // Open QR modal by setting URI.
+        setWalletConnectUri(uri);
```
wgwz, line 35: could you explain a bit more about what this URI is and how it is working? this might be a good spot in the code to some comments documenting this a little bit.

web-registry/src/components/organisms/WalletButton/WalletButton.tsx
```diff
@@ -0,0 +1,79 @@
...
+        {error && (
+          <ErrorBanner text="Please install Keplr extension to use Regen Ledger features" />
+        )}
+      </div>
+      <WalletModal
+        open={isModalOpen}
+        onClose={onModalClose}
+        wallets={walletsUiConfig}
+        state={modalState}
+        uri={walletConnectUri}
```
wgwz, line 70: maybe a name change here could be an improvement? maybe it could slightly lower the mental bandwidth to understand what this URI is for? i.e. `uri` => `qrCodeURI`

web-registry/src/lib/wallet/wallet.utils.ts
```diff
@@ -0,0 +1,85 @@
...
+  setWalletConnectUri: UseStateSetter<string | undefined>;
+  onQrCloseCallback: MutableRefObject<(() => void) | undefined>;
+};
+
+export const getWalletConnectInstance = async ({
+  setWalletConnectUri,
+  onQrCloseCallback,
+}: GetWalletConnectInstanceParams): Promise<WalletConnect> => {
+  const walletConnect = new (await import('@walletconnect/client')).default({
+    bridge: 'https://bridge.walletconnect.org',
```
wgwz, line 27: should we move this into a constants file? same for the signing methods below?

web-registry/src/lib/wallet/hooks/useAutoConnect.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { WalletType } from '../walletsConfig/walletsConfig.types';
+import { ConnectWalletType } from './useConnectWallet';
+
+type Props = {
+  setError: UseStateSetter<unknown>;
+  setLoaded: UseStateSetter<boolean>;
+  connectWallet: ConnectWalletType;
+};
+
+export const useAutoConnect = ({
```
wgwz, line 15: could we add some comments here describing the intended purpose/use-case? i.e. it's not immediately obvious to me why we want to have an autoconnect?

web-registry/src/lib/wallet/hooks/useDetectKeplrMobileBrowser.tsx
```diff
@@ -0,0 +1,34 @@
...
+  useEffect(() => {
+    if (typeof window === 'undefined' || !loaded || wallet.address !== '') {
+      return;
+    }
+
+    import('@keplr-wallet/stores')
+      .then(({ getKeplrFromWindow }) => getKeplrFromWindow())
+      .then(keplr => {
+        if (keplr && keplr.mode === 'mobile-web') {
+          connectWallet({ walletType: WalletType.Keplr });
```
wgwz, line 29: this me just misunderstanding something and looking for clarity. why do we need `WalletType.Keplr` here and not `WalletType.WalletConnectKeplr`? it seems like we're in mobile, so shouldn't we always be using wallet connect in this case?

web-registry/src/lib/wallet/wallet.utils.ts
wgwz: also curious about what this `callback` does, it seems like it later becomes the `onQrCloseCallback`. maybe a rename here would be helpful for consistency?

web-registry/src/lib/wallet/hooks/useConnectWallet.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+      const walletClient = await walletConfig?.getClient({
+        chainInfo,
+        walletConnect,
+      });
+
+      // Only Keplr browser extension supports suggesting chain.
+      // Not WalletConnect nor embedded Keplr Mobile web.
+      if (
+        walletConfig?.type === WalletType.Keplr &&
```
wgwz, line 65:
```suggestion
        isKeplr &&
```

web-registry/src/lib/wallet/hooks/useWalletConnectFinalize.tsx
```diff
@@ -0,0 +1,46 @@
...
+  useEffect(() => {
+    const onWalletConnectEvent = async (): Promise<void> => {
+      const walletClient = await walletConfigRef.current?.getClient({
+        chainInfo,
+        walletConnect,
+      });
+
+      finalizeConnection({
+        setWallet,
+        walletClient: walletClient,
```
wgwz, line 31:
```suggestion
        walletClient,
```

web-registry/src/lib/wallet/wallet.tsx
```diff
@@ -0,0 +1,106 @@
...
+    walletConfigRef,
+  });
+
+  const connect = useConnect({ connectWallet, setConnectionType, setError });
+
+  useAutoConnect({ connectWallet, setError, setLoaded });
+  useOnAccountChange({ connectWallet, wallet });
+  useDetectKeplrMobileBrowser({ connectWallet, loaded, wallet });
+  useWalletConnectCallback({ onQrCloseCallback, walletConnectUri });
+  useWalletConnectFinalize({ setWallet, walletConfigRef, walletConnect });
```
wgwz, line 86: one thing i'm unsure of that would be nice to understand is why we have this separate finalize action, and why we need this one, when we also have a finalize action inside of `useConnectWallet`.

web-registry/src/components/organisms/WalletButton/WalletButton.utils.ts
```diff
@@ -0,0 +1,51 @@
...
+type GetMobileConnectUrlParams = {
+  uri?: string;
+};
+
+export const getMobileConnectUrl = ({
+  uri,
+}: GetMobileConnectUrlParams): string | undefined => {
+  if (checkIsMobile() && uri) {
+    return checkIsAndroid()
+      ? `intent://wcV1?${uri}#Intent;package=com.chainapsis.keplr;scheme=keplrwallet;end;`
```
wgwz, line 46: maybe a comment to docs about the format of these?

web-registry/src/hooks/useQuerySellOrders.ts
```diff
@@ -29,15 +29,16 @@ export const useQuerySellOrders = function (): {
...
     QuerySellOrdersResponseExtented | undefined
   >(undefined);
 
   useEffect(() => {
-    if (!api?.queryClient) return;
+    if (!api?.queryClient || queryClient !== undefined) return;
     const _queryClient: QueryClientImpl = new QueryClientImpl(api.queryClient);
     setQueryClient(_queryClient);
-  }, [api?.queryClient]);
+  }, [api?.queryClient, queryClient]);
```
wgwz, line 41: just curious to know what this fixed?

wgwz (approved):
Awesome work @flagrede! Just added a few small suggestions and mostly asked some questions to understand it better. Thanks!

erikalogie:
> @erikalogie I've setup a sandbox to deploy with production configuration: https://alike-spring.surge.sh/ You should be able able to test the mobile connection now. I've also handles your feedback on this deploy.

@flagrede sorry I missed this notification somehow! I just tested this on desktop and mobile. The desktop connection is still not working, I'm getting the same "this site can't be reached" error.

Mobile is working great, one piece of feedback, it would be great to truncate the address, right now it appears as such: 
Screenshot_20221028-065215. Originally I had intended to not have the address at all on mobile but maybe we can leave it for now as long as it is sufficiently truncated.

flagrede:
@wgwz thanks for all the feedback! Will handle them on Monday.

@erikalogie Did you scan the QRCode with Keplr mobile app or another QRCode scanner app ? (the QRCode is meant to be scanned with the Keplr app only).
I'm having no issue connecting with the QR on desktop so I'm wondering if it might be that.
I'll truncate the address, sounds better indeed.

erikalogie:
@flagrede I was doing it with the regular QR code reader (whoops!) but it's also not working for me with the app, it just has a loading sign that never goes away on the qr code scan screen. 
Screenshot_20221028-115237

Either way, I think that we should add something about this to the instructions, so changing the title to "Scan with the Keplr mobile app".

web-components/src/components/modal/wallet-modal/components/WalletModal.Mobile.config.tsx
flagrede: Completely agree with that. Had to use it here because it's an array definition and not a function.
I think I'll move this logic to the filter part to make it more clear.

web-registry/package.json
flagrede: Add to use that to fix an alias issue. This script allows us to override the webpack alias config, you can see it here: https://github.com/regen-network/regen-web/pull/1514/files#diff-7948d3454e9336c2f9cccf4385651c122ee3940f1b4403030971e1d63fe609dd

web-components/src/components/modal/wallet-modal/components/WalletModal.Mobile.config.tsx
flagrede: I'm confused, isn't it the same file that you commented on here: https://github.com/regen-network/regen-web/pull/1514#discussion_r1007304128 ?

web-registry/src/lib/wallet/wallet.utils.ts
flagrede: Good point, I added the related explanation from their doc here:
```
// The protocol establishes a remote connection between two apps and/or devices using a Bridge server to relay payloads.
// These payloads are symmetrically encrypted through a shared key between the two peers.
// The connection is initiated by one peer displaying a QR Code or deep link with a standard WalletConnect URI and is established when the counter-party approves this connection request.
```

flagrede: `onQrCloseCallback ` is a callback to clean the URI within wallet connect once the connection has been established.
Agree with the naming, will make it more explicit.

web-registry/src/lib/wallet/hooks/useAutoConnect.tsx
flagrede: Well, it's a behavior we already had, the idea is to auto-connect the user if he already made a connection before and did not disconnect.
Will add some comments.

web-registry/src/lib/wallet/hooks/useOnAccountChange.tsx
```diff
@@ -0,0 +1,30 @@
...
+    const listener = (): void => {
+      if (wallet) {
+        connectWallet({ walletType: WalletType.Keplr });
+      }
+    };
+
+    window.addEventListener('keplr_keystorechange', listener);
+
+    return () => {
+      window.removeEventListener('keplr_keystorechange', listener);
```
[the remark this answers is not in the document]
flagrede: Listeners are meant to be removed when they are no longer needed because they use resources.

web-registry/src/lib/wallet/hooks/useDetectKeplrMobileBrowser.tsx
flagrede: This is intuitive to understand indeed. Basically, when the app will be used within Keplr native app (with their mobile browser) the connection settings will be the same as using the chrome extension.
Will add some comments to make that clear.

web-registry/src/lib/wallet/wallet.tsx
flagrede: We need a separate effect for the wallet connect finalize action because it happens from a listener event (while on the desktop we can do the finalize directly).

web-registry/src/hooks/useQuerySellOrders.ts
flagrede: We had an infinite loop here because the client was redefined on each render.

flagrede:
@wgwz thanks again for the review, I should have handled/responded to all your feedback. Let me know if they are still some unclear areas.

@erikalogie The connection with a QRCode should be instant. I made some more tests and I'm always able to connect with it. Did you have the notification on the mobile app when you scanned the QRCode?

erikalogie:
@flagrede it is working as expected now, the connection was instant

web-components/src/components/modal/wallet-modal/components/WalletModal.Mobile.config.tsx
wgwz: whoops! sorry about that, i got lost in the code somehow 😆

web-registry/src/lib/wallet/hooks/useAutoConnect.tsx
wgwz: thanks for the doc string there, that helps 👍

web-registry/src/lib/wallet/hooks/useOnAccountChange.tsx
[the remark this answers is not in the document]
wgwz: So in this case, is the parent component the `WalletProvider`? In terms of the app, is there a particular action/s when that would get unmounted? thanks for taking the time to explain these things while i learn some react 🙏

web-components/src/components/modal/wallet-modal/components/WalletModal.Mobile.config.tsx
wgwz: thanks for changing this, i think it is more readable with the filter

web-registry/src/lib/wallet/hooks/useOnAccountChange.tsx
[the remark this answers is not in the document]
flagrede: I agree here this should never happen since `WalletProdiver` is in the root app tree. That being said it's best to keep the cleaning function, it might be useful for example if the hook is used elsewhere later on. We might even have a lint for writing those cleaning functions.

victordelval:
I've also tested it and it works fine!

victordelval:
@erikalogie @flagrede one question, I just realized that in mobile we are representing the user's address by abbreviating it differently than the standard way... we should change it, right?

Mobile
`regen1df675r9vnf7pdedn4sf26...`

Standard
`regen1df...gxmy46`

erikalogie:
> @erikalogie @flagrede one question, I just realized that in mobile we are representing the user's address by abbreviating it differently than the standard way... we should change it, right?
> 
> Mobile `regen1df675r9vnf7pdedn4sf26...`
> 
> Standard `regen1df...gxmy46`

Yes, let's fix this, great catch

blushi:
I could connect successfully both from the app on desktop using wallet connect with QR code and from mobile directly through the Keplr mobile app.
From mobile, I tried to purchase some credits but then when I hit "purchase", the form just becomes disabled and nothing else happens:

IMG_9459F66CF1D1-1

blushi:
> I could connect successfully both from the app on desktop using wallet connect with QR code and from mobile directly through the Keplr mobile app. From mobile, I tried to purchase some credits but then when I hit "purchase", the form just becomes disabled and nothing else happens:
> 
> IMG_9459F66CF1D1-1

Ok actually you have to go manually to the Keplr app and then you can approve from there. But we should probably prompt the user to do so.

victordelval (approved):
Incredible work @flagrede ! Many thanks! The review has been much easier thanks to the introduction you did the other day and @wgwz 's questions!

I only have a couple of questions, not so much to change now, but to discuss and reach consensus and unify criteria.

* Type aliases vs Interfaces (official documentation defaults to using interfaces when indistinct, [mainly because errors messages are better](https://www.typescriptlang.org/play?e=83#example/types-vs-interfaces))
* Guard clause style (mainly in effect hooks, looks more clear and avoids nesting)

blushi (commented):
Great work, thanks for refactoring and cleaning up!
Just a few comments UX wise.

web-components/src/components/modal/wallet-modal/components/WalletModal.Mobile.config.tsx
```diff
@@ -0,0 +1,46 @@
...
+      ),
+    },
+    {
+      label: 'Scan with the Keplr mobile app',
+      content: (
+        <Center sx={{ pt: 9, height: 340 }}>
+          <QRCode size={300} value={qrCodeUri ?? ''} />
+        </Center>
+      ),
+    },
```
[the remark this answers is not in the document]
flagrede: Actually, I mirrored Osmosis UX here. I think the idea is that it should be possible for example to connect from iPad while having the wallet only on a phone.
In any case, we always try to open the deep link by default, so the QR code on mobile will only be visible if the user is on a mobile device with no wallet installed.

flagrede:
> > I could connect successfully both from the app on desktop using wallet connect with QR code and from mobile directly through the Keplr mobile app. From mobile, I tried to purchase some credits but then when I hit "purchase", the form just becomes disabled and nothing else happens
> 
> Ok actually you have to go manually to the Keplr app and then you can approve from there. But we should probably prompt the user to do so.

@erikalogie Wdyt, should we make a screen for that ?

haveanicedavid (approved):
Great work

web-registry/src/lib/wallet/chainInfo/chainInfo.ts
```diff
@@ -0,0 +1,87 @@
...
+    {
+      // Coin denomination to be displayed to the user.
+      coinDenom: 'REGEN',
+      // Actual denom (i.e. uatom, uscrt) used by the blockchain.
+      coinMinimalDenom: 'uregen',
+      // # of decimal points to convert minimal denomination to user-facing denomination.
+      coinDecimals: 6,
+      // (Optional) Keplr can show the fiat value of the coin if a coingecko id is provided.
+      // You can get id from https://api.coingecko.com/api/v3/coins/list if it is listed.
+      // coinGeckoId: ""
```
mhagel, line 55: we might have this now?

web-registry/src/lib/wallet/wallet.constants.ts
```diff
@@ -0,0 +1,16 @@
+export const AUTO_CONNECT_WALLET_KEY = 'auto_connect_wallet';
+export const WALLET_CONNECT_KEY = 'walletconnect';
+export const WALLET_CONNECT_BRIDGE_URL = 'https://bridge.walletconnect.org';
+export const WALLET_CONNECT_SIGNING_METHODS = [
+  'keplr_enable_wallet_connect_v1',
+  'keplr_sign_amino_wallet_connect_v1',
+];
+
+export const walletConnectClientMeta = {
+  name: 'Regen Network Registry',
```
mhagel, line 10:
```suggestion
  name: 'Regen Marketplace',
```
We are using "marketplace" instead of "registry" now

mhagel (approved):
This is great, thank you. Left a couple minor comments.

web-components/src/components/modal/wallet-modal/components/WalletModal.Mobile.config.tsx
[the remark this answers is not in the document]
blushi: > In any case, we always try to open the deep link by default, so the QR code on mobile will only be visible if the user is on a mobile device with no wallet installed. 

I had Keplr mobile app installed and even if it tried to open the deep link by default indeed, I could cancel that and go to the "Scan with the Keplr mobile app" tab.

> Actually, I mirrored Osmosis UX here. I think the idea is that it should be possible for example to connect from a tablet while having the wallet only on a phone.

I then tried to scan the QR code using a second phone with Keplr app on it but that wasn't successful : once I scanned the QR code from the app, I could see the loading spinner for some time then it just disappeared but nothing else happened.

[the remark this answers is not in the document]
flagrede: @blushi I removed the QR code on mobile since it's not really working and it might have been a weird case, to begin with.
I redeployed the sandbox env, if you want to run some final tests: https://alike-spring.surge.sh/

flagrede:

PR #1551, opened 2022-11-09
feat: Add requested metadata to buy, send, and sell events

wgwz:
## Description

Closes: https://github.com/regen-network/regen-registry/issues/1341

---

### Author Checklist

I have...

#### Setting up backport PRs

   your PR:

   ```
   git merge dev
   git merge dev
   ```

### How to test

1. Go to the deploy preview: https://deploy-preview-1551--regen-registry.netlify.app/
2. Open the "Regen Marketplace [DEV]" GA dashboard (realtime tab): https://analytics.google.com/analytics/web/#/p340668480/realtime/overview?params=_u..nav%3Dmaui
3. Complete all actions in the tracking spreadsheet: https://docs.google.com/spreadsheets/d/1JXw_pDC4QyqtcdYgxpMkbTVc24E2yz-qpwxvzelPfnY/edit?usp=sharing
    a. At each step, check that you see the corresponding event in GA or Amplitude.

### Reviewers Checklist

I have...

- [ ] reviewed tests

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -20,8 +20,6 @@
   REACT_APP_BLOCK_EXPLORER = "https://redwood.regen.aneka.io"
   REACT_APP_SENTRY_ENVIRONMENT = "development"
   REACT_APP_BRIDGE_CREDIT_CLASS_ID = "C46"
-  # TODO: remove before merging
-  REACT_APP_ANALYTICS_ENABLED="true"
 
 [context.hambach.environment]
   REACT_APP_SANITY_TAG = "default"
--- a/netlify.toml
+++ b/netlify.toml
@@ -20,6 +20,8 @@
   REACT_APP_BLOCK_EXPLORER = "https://redwood.regen.aneka.io"
   REACT_APP_SENTRY_ENVIRONMENT = "development"
   REACT_APP_BRIDGE_CREDIT_CLASS_ID = "C46"
+  # TODO: remove before merging
+  REACT_APP_ANALYTICS_ENABLED="true"
 
 [context.hambach.environment]
   REACT_APP_SANITY_TAG = "default"
--- a/web-components/src/components/cards/ProjectCard/ProjectCard.tsx
+++ b/web-components/src/components/cards/ProjectCard/ProjectCard.tsx
@@ -2,7 +2,8 @@ import { useState } from 'react';
 import { useLocation } from 'react-router-dom';
 import { Box, SxProps, Theme, useTheme } from '@mui/material';
 import clsx from 'clsx';
-import { useAnalytics } from 'use-analytics';
+import { Buy1Event } from 'web-registry/src/lib/tracker/types';
+import { useTracker } from 'web-registry/src/lib/tracker/useTracker';
 
 import { formatStandardInfo } from '../../../utils/format';
 import OutlinedButton from '../../buttons/OutlinedButton';
@@ -61,7 +62,7 @@ export function ProjectCard({
   const theme = useTheme();
   const { classes } = useProjectCardStyles();
   const location = useLocation();
-  const { track } = useAnalytics();
+  const { track } = useTracker();
 
   const [open, setOpen] = useState<boolean>(true);
 
@@ -230,13 +231,14 @@ export function ProjectCard({
                   <OutlinedButton
                     onClick={event => {
                       event.stopPropagation();
-                      track('buy1', {
+                      const trackData: Buy1Event = {
                         url: location.pathname,
                         cardType: 'project',
                         buttonLocation: 'projectCard',
                         projectName: name,
                         projectId: id,
-                      });
+                      };
+                      track('buy1', trackData);
                       onButtonClick && onButtonClick();
                     }}
                     size="small"
--- a/web-components/src/components/cards/ProjectCard/ProjectCard.tsx
+++ b/web-components/src/components/cards/ProjectCard/ProjectCard.tsx
@@ -233,15 +233,14 @@ export function ProjectCard({
                   <OutlinedButton
                     onClick={event => {
                       event.stopPropagation();
-                      const trackData: Buy1Event = {
+                      track<'buy1', Buy1Event>('buy1', {
                         url: location.pathname,
                         cardType: 'project',
                         buttonLocation: 'projectCard',
                         projectName: name,
                         projectId: id,
                         creditClassId,
-                      };
-                      track('buy1', trackData);
+                      });
                       onButtonClick && onButtonClick();
                     }}
                     size="small"
--- a/web-components/src/components/cards/ProjectCard/ProjectCard.tsx
+++ b/web-components/src/components/cards/ProjectCard/ProjectCard.tsx
@@ -22,6 +22,7 @@ import { getAbbreviation } from './ProjectCard.utils';
 export interface ProjectCardProps extends MediaCardProps {
   id?: string;
   name: string;
+  creditClassId?: string;
   imgSrc: string;
   place: string;
   area: number;
@@ -42,6 +43,7 @@ export interface ProjectCardProps extends MediaCardProps {
 export function ProjectCard({
   id,
   name,
+  creditClassId,
   imgSrc,
   place,
   area,
@@ -237,6 +239,7 @@ export function ProjectCard({
                         buttonLocation: 'projectCard',
                         projectName: name,
                         projectId: id,
+                        creditClassId,
                       };
                       track('buy1', trackData);
                       onButtonClick && onButtonClick();
--- a/web-components/src/components/form/CreditSendForm.tsx
+++ b/web-components/src/components/form/CreditSendForm.tsx
@@ -129,7 +129,6 @@ const CreditSendForm: React.FC<React.PropsWithChildren<FormProps>> = ({
       initialValues={initialValues}
       validate={validateHandler}
       onSubmit={async values => {
-        track<'send2'>('send2'); //TODO: move to onSubmit, regen-network/regen-registry#1365
         onSubmit(values);
       }}
     >
--- a/web-components/src/components/form/CreditSendForm.tsx
+++ b/web-components/src/components/form/CreditSendForm.tsx
@@ -2,7 +2,6 @@ import React from 'react';
 import { Box } from '@mui/system';
 import { Field, Form, Formik, FormikErrors } from 'formik';
 import { makeStyles } from 'tss-react/mui';
-import { useTracker } from 'web-registry/src/lib/tracker/useTracker';
 
 import { Theme } from '../../theme/muiTheme';
 import { Flex } from '../box';
@@ -83,7 +82,6 @@ const CreditSendForm: React.FC<React.PropsWithChildren<FormProps>> = ({
   onSubmit,
 }) => {
   const { classes: styles } = useStyles();
-  const { track } = useTracker();
 
   const initialValues = {
     sender,
--- a/web-components/src/components/form/CreditSendForm.tsx
+++ b/web-components/src/components/form/CreditSendForm.tsx
@@ -129,7 +129,7 @@ const CreditSendForm: React.FC<React.PropsWithChildren<FormProps>> = ({
       initialValues={initialValues}
       validate={validateHandler}
       onSubmit={async values => {
-        track('send2'); //TODO: move to onSubmit, regen-network/regen-registry#1365
+        track<'send2'>('send2'); //TODO: move to onSubmit, regen-network/regen-registry#1365
         onSubmit(values);
       }}
     >
--- a/web-components/src/components/form/CreditSendForm.tsx
+++ b/web-components/src/components/form/CreditSendForm.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { Box } from '@mui/system';
 import { Field, Form, Formik, FormikErrors } from 'formik';
 import { makeStyles } from 'tss-react/mui';
-import { useTrack } from 'use-analytics';
+import { useTracker } from 'web-registry/src/lib/tracker/useTracker';
 
 import { Theme } from '../../theme/muiTheme';
 import { Flex } from '../box';
@@ -83,7 +83,7 @@ const CreditSendForm: React.FC<React.PropsWithChildren<FormProps>> = ({
   onSubmit,
 }) => {
   const { classes: styles } = useStyles();
-  const track = useTrack();
+  const { track } = useTracker();
 
   const initialValues = {
     sender,
--- a/web-registry/src/lib/tracker/types.ts
+++ b/web-registry/src/lib/tracker/types.ts
@@ -1,0 +1,25 @@
+export interface LoginEvent {
+  date: string;
+  account: string;
+}
+
+export interface Buy1Event {
+  url: string;
+  cardType?: string;
+  buttonLocation: string;
+  projectName?: string | null;
+  projectId?: string | null;
+  creditClassName?: string;
+  creditClassId?: string;
+}
+
+export interface Buy2Event {
+  url: string;
+  price?: string;
+  batchDenom?: string;
+  projectName?: string | null;
+  creditClassId?: string;
+  quantity?: number;
+  currencyDenom?: string;
+  retirementAction?: string;
+}
--- a/web-registry/src/lib/tracker/types.ts
+++ b/web-registry/src/lib/tracker/types.ts
@@ -18,6 +18,7 @@ export interface Buy2Event {
   price?: string;
   batchDenom?: string;
   projectName?: string | null;
+  projectId?: string | null;
   creditClassId?: string;
   quantity?: number;
   currencyDenom?: string;
--- a/web-registry/src/lib/tracker/types.ts
+++ b/web-registry/src/lib/tracker/types.ts
@@ -10,7 +10,7 @@ export interface Buy1Event {
   projectName?: string | null;
   projectId?: string | null;
   creditClassName?: string;
-  creditClassId?: string;
+  creditClassId?: string | null;
 }
 
 export interface Buy2Event {
--- a/web-registry/src/lib/tracker/types.ts
+++ b/web-registry/src/lib/tracker/types.ts
@@ -84,14 +84,31 @@ export interface SellFailureEvent extends SellOutcomeBaseEvent {
   errorMessage?: string;
 }
 
-export interface Send1Event {
+interface SendBaseEvent {
+  batchDenom: string;
+  creditClassId?: string;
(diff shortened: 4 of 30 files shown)


web-registry/src/lib/tracker/useTracker.ts
```diff
@@ -0,0 +1,17 @@
...
+import { Buy2Event, LoginEvent } from './types';
+
+interface UseTracker {
+  track: (
+    eventName: string,
+    payload?: any | LoginEvent | Buy2Event,
+    options?: any,
+    callback?: (...params: any[]) => any,
+  ) => Promise<any>;
+}
```
victordelval, line 12: @wgwz some comments on this interface. Here you are mimicking the original interface of the library, but that is too open (unspecific) for our use cases.

I think that at the moment we are compulsorily informing the first two parameters (event name and payload). Therefore, they should be the only properties of the track function that we are wrapping at the moment, and the two mandatory ones.

The first, eventName, could be a type of the following form:
```
type EventName = 'login' | 'buy1' | 'buy2';
```
The Payload could be as follows:
```
type EventPayload = LoginEvent | Buy1Event | Buy2Event;
```
Finally, we would have a restrictive interface, where any name or payload event is not valid:
```
interface UseTracker {
   tracks: (
     eventName: EventName,
     payload: EventPayload
   ) => Promise<any>;
}
```

web-registry/src/lib/tracker/types.ts
```diff
@@ -0,0 +1,26 @@
...
+  url: string;
+  price?: string;
+  batchDenom?: string;
+  projectName?: string | null;
+  projectId?: string | null;
+  creditClassId?: string;
+  quantity?: number;
+  currencyDenom?: string;
+  retirementAction?: string;
+}
```
victordelval, line 26: One comment about these interfaces for the types of events we are going to track. If we add a static property to each of these event types, such as `eventType`, or to continue with `eventName`, like so:

When we build the data to track, as soon as we set this static property, then the rest of the properties that correspond to the type are **inferred**. This way there is no need to import the types at the place where the track function is used.

```
export interface LoginEvent {
  eventName: 'login';                 <<<< Here, for the magic!
  date: string;
  account: string;
}

export interface Buy1Event {
  eventName: 'buy1';                  <<<< Here, for the magic!
  url: string;
  cardType?: string;
  buttonLocation: string;
  projectName?: string | null;
  projectId?: string | null;
  creditClassName?: string;
  creditClassId?: string | null;
}

export interface Buy2Event {
  eventName: 'buy2';                  <<<< Here, for the magic!
  url: string;
  price?: string;
  batchDenom?: string;
  projectName?: string | null;
  projectId?: string | null;
  creditClassId?: string;
  quantity?: number;
  currencyDenom?: string;
  retirementAction?: string;
}
```

If to this improvement, we add another twist to the `useTracker` hook, we can have the following implementation that is very **powerful** (a single input `EventData` that is the union of `EventName` and `EventPayload`)

```
// lib/tracker/types.ts

export type EventData = LoginEvent | Buy1Event | Buy2Event;
```
```
// lib/tracker/useTracker.ts

import { useCallback } from 'react';
import { useTrack } from 'use-analytics';

import { EventData } from './types';

interface UseTracker {
  track: (eventData: EventData) => void;
}

export const useTracker = (): UseTracker => {
  const track = useTrack();

  // our custom tracker function that we expose...
  const tracker = useCallback(
    (eventData: EventData) => {
      const { eventName, ...restEventData } = eventData;
      track(eventName, restEventData);
    },
    [track],
  );

  return {
    track: tracker;
  };
};

```

victordelval: @wgwz the ideal would be to link `eventName` and `eventPayload` in the same data structure, so by reporting only the `eventName` property, the TS itself would infer the corresponding type and there is no possibility of passing the part of the payload that does not correspond to the type. 

Right now as it is, with separate `eventName` and `eventPayload`, nothing prevents us from passing a payload that doesn't correspond to the event type, something like: `track('login', dataBuy1Event)`

wgwz: i think i might prefer an approach that uses a generic, i.e.: 

```typescript
function tracker<T>(eventName: string, eventData: T): T {
 return eventData;
}

interface LoginEvent {
    account: string;
}

const result = tracker<LoginEvent>('login', {account: "regen12345"});

console.log(result);
```

i don't like stuffing the event name into the event data. i prefer having the event name as the first argument to the function. i think it is a distinct and important argument to this function that having it's own position leads to more readable code.

with a generic we will still get type safety: 

Screen Shot 2022-11-15 at 10 14 34 AM

victordelval: @wgwz mm, I understand your proposal and it is an option, but I think that the other proposal could be more powerful because it is based on **discriminated unions**, which is a powerful technique... I think that in this case we don't need to create a generic function, but our custom wrapper, which is very specific since it expects our custom event type data that the function itself should know beforehand (its not a generic modulo, but our custom tracker). Creating event types that combine `eventName` (understood as `type` or `kind` property, more idiomatic) with the specific data of the event type (`eventPayload`) is not stuffing the data structures at all!... this allows discriminated unions (related to algebraic types, set theory, etc), and in the code editor type narrowing when using the function...

* https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions
* https://blog.logrocket.com/understanding-discriminated-union-intersection-types-typescript/#discriminated-union

@blushi @flagrede @haveanicedavid @mhagel any thoughts on this?

victordelval: @wgwz also with that approach with generics i think that nothing keep you safe from doing something like this, mixing buy1 name and loginData:

```
tracker<LoginEvent>('buy1', loginData);
```

mhagel: I like when code is easily searchable. So, if I want to find what line in the code has the 'buy1' event, I can search for it. It sounds like this could be accomplished either way.

wgwz: @victordelval we can protect ourselves from that situation by doing it like this:

```typescript
function tracker<S, T>(eventName: S, eventData: T): T {
 return eventData;
}

interface LoginEvent {
    account: string;
}

const result = tracker<'login', LoginEvent>('login', {});

console.log(result);
```

web-components/src/components/cards/ProjectCard/ProjectCard.tsx
```diff
@@ -230,13 +233,15 @@ export function ProjectCard({
...
                         url: location.pathname,
                         cardType: 'project',
                         buttonLocation: 'projectCard',
                         projectName: name,
                         projectId: id,
-                      });
+                        creditClassId,
+                      };
+                      track('buy1', trackData);
                       onButtonClick && onButtonClick();
```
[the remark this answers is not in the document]
mhagel: I'm ok with this being a separate effort though

web-components/src/components/form/CreditSendForm.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { makeStyles } from '@mui/styles';
 import { Box } from '@mui/system';
 import { Field, Form, Formik, FormikErrors } from 'formik';
-import { useTrack } from 'use-analytics';
+import { useTracker } from 'web-registry/src/lib/tracker/useTracker';
```
mhagel, line 5: again, it feels like a step in the wrong direction importing web-registry modules into web-components. Prefer to keep all tracking code in the codebase of the app we are tracking. 

Another way we could do this is to simply wrap a web-component in a new web-registry component, and add the tracking logic there. That would mean less repeated code than my previous suggestion.

web-components/src/components/cards/ProjectCard/ProjectCard.tsx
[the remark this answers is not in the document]
wgwz: is there an alternative we could explore where we only make a call to the track function if it has been passed to the component?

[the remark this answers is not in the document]
wgwz: the idea being that it's only there as an optional dependency in that scenario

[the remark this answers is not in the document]
wgwz: opened a new issue for resolving this separately: https://github.com/regen-network/regen-registry/issues/1400

web-registry/src/lib/tracker/types.ts
mhagel: There are some interesting ideas here, but maybe we are at risk of over-engineering this for a simple tracking use case. 

Personally, as a dev, I would expect to be able to add a tracking event wherever I want, with a custom payload each time. I'm not sure I follow everything that was suggested here, but that is the key goal of this task.

Maybe we should get this done with a simple implementation, and remember this conversation for future cases. Or we could make a new issue for type-safety.

@wgwz @victordelval

web-registry/src/pages/Marketplace/Storefront/Storefront.normalizer.ts
```diff
@@ -58,7 +58,9 @@ export const normalizeProjectsInfosByHandleMap = ({
 
     projectsMap.set(project?.id, {
       name: project.metadata?.['schema:name'] || project.id,
+      // TODO: why is this being called classIdName?
       classIdName: creditClassSanity?.nameRaw ?? project?.classId,
+      // TODO: why is this being called classIdUrl?
```
flagrede, line 63: Good catch, definitely something weird here.

flagrede (commented):
After catching up on the tracker module discussion, I'm fine with the current implementation.

web-registry/src/pages/Dashboard/MyEcocredits/MyEcocredits.tsx
```diff
@@ -291,7 +291,7 @@ export const MyEcocredits = (): JSX.Element => {
                       icon: <AvailableCreditsIconAlt sx={sxs.arrow} />,
                       label: CREATE_SELL_ORDER_SHORT,
                       onClick: () => {
-                        track('sell1');
+                        track<'sell1'>('sell1');
```
blushi, line 294: why aren't we adding metadata here?

web-registry/src/pages/Dashboard/MyEcocredits/MyEcocredits.tsx
```diff
@@ -305,7 +305,7 @@ export const MyEcocredits = (): JSX.Element => {
                       ),
                       label: CREDIT_SEND_TITLE,
                       onClick: () => {
-                        track('send1');
+                        track<'send1'>('send1');
```
blushi, line 308: ditto

web-registry/src/pages/Marketplace/Storefront/Storefront.normalizer.ts
```diff
@@ -58,7 +58,9 @@ export const normalizeProjectsInfosByHandleMap = ({
 
     projectsMap.set(project?.id, {
       name: project.metadata?.['schema:name'] || project.id,
+      // TODO: why is this being called classIdName?
```
blushi, line 61: if there's some credit class name, then we display it, else we just display the credit class id
maybe `classIdOrName` would be more explicit but sounds a bit long too

web-registry/src/pages/Marketplace/Storefront/Storefront.normalizer.ts
blushi: looking at its usage, I think it's because it's meant to be used within URLs but I guess `classId` would just be ok too...

web-registry/src/lib/tracker/types.ts
blushi: yeah I see pros and cons for both approaches... although I have a preference for @victordelval's approach. Since we do know the events we're expecting, I think it's best practice to have to define a new interface in order to support a new event type. We could use `$type` instead of `eventName`.
Also the "generics" approach makes the code a bit repetitive: `tracker<'login', LoginEvent>('login', { ... });` and it doesn't enforce the association of a particular event type with the event payload, you could still do `tracker<'buy1', LoginEvent>('buy1', { ... });` for instance.

web-registry/src/pages/Dashboard/MyEcocredits/MyEcocredits.tsx
wgwz: same here, i have a commit with this that i haven't pushed yet

web-registry/src/pages/Dashboard/MyEcocredits/MyEcocredits.tsx
wgwz: i have a commit with this that i haven't pushed yet

web-registry/src/lib/tracker/types.ts
```diff
@@ -0,0 +1,54 @@
...
+  creditClassId?: string | null;
+}
+
+export interface Buy2Event {
+  url: string;
+  price?: string;
+  batchDenom?: string;
+  projectName?: string | null;
+  projectId?: string | null;
+  creditClassId?: string;
```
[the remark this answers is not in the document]
wgwz: there are some call sites for the `track` function where the credit class name is not readily available, and where i find it difficult to figure out how to efficiently pass it to. open to suggestions on how to improve that for these cases where i did not include it. but since we do have `projectId` and `creditClassId` available in almost all places, i almost wonder if we can just adjust the expectation in our analytics so that we don't include the credit class name. it seems to me like the ID is the more important field anyways, from an analytics perspective. it's a simpler key to aggregate on, etc.

my concern on the technical side was that i didn't want to add an api call to lookup the name prior to sending the tracking event, and introduce a performance issue.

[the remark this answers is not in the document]
blushi: yeah I agree just including the (on chain) ids would be sufficient
any thoughts @erikalogie ?

web-registry/src/lib/tracker/types.ts
wgwz: > it doesn't enforce the association of a particular event type with the event payload, you could still do tracker<'buy1', LoginEvent>('buy1', { ... }); for instance.

i could do the same thing if `LoginEvent` included `eventName`, it would just require modifications across different files, rather than at the call site. i'd argue it's a version of "security by obscurity". maybe i am missing something though.

> We could use $type instead of eventName

@blushi could you explain what the $type is here?

blushi: > @blushi could you explain what the $type is here?

just a suggestion for a different key name, similarly to what we have already in the @regen-network/api generated types

netlify.toml
```diff
@@ -18,6 +18,8 @@
   REACT_APP_LEDGER_CHAIN_NAME = "Regen - Redwood Testnet"
   REACT_APP_BLOCK_EXPLORER = "https://redwood.regen.aneka.io"
   REACT_APP_SENTRY_ENVIRONMENT="development"
+  # TODO: remove before merging
```
wgwz, line 21: TODO before merge

web-registry/src/lib/tracker/types.ts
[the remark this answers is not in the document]
wgwz: Took care of this item @victordelval , please take a look if you want, I will mark this as resolved now

web-components/src/components/cards/ProjectCard/ProjectCard.tsx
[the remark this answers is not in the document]
wgwz: @blushi glad to clean that up in the upcoming sprint, it shouldn't be too hard to do.

web-registry/src/pages/Marketplace/Storefront/Storefront.normalizer.ts
wgwz: i personally think that would be better, because when i saw this data in a downstream component, i forget where i was, the name definitely threw me off when i was inspecting the object.

wgwz: should we address this now if it's a small change? otherwise create a new issue?

web-registry/src/lib/tracker/types.ts
wgwz: i think given the feedback, i would like to merge this with the generics approach. we can create a separate issue if refactoring and using the approach that @victordelval outlined becomes favorable in the future. i just finished manually testing the added metadata to these events and it's working as intended. would prefer to wait for more changes and another round of testing, in favor getting this deployed, since it does have some important updates for @erikalogie and the marketing team in analytics tools. i.e. error messages when we have buy failures!

web-registry/src/pages/Marketplace/Storefront/Storefront.normalizer.ts
wgwz: did a find and replace in this commit: 5bf681b055666e18ada4556ed072002aa37f0040

web-registry/src/pages/Marketplace/Storefront/Storefront.normalizer.ts
blushi: thanks!

web-registry/src/pages/Marketplace/Storefront/Storefront.normalizer.ts
blushi: thanks, resolving

web-registry/src/lib/tracker/types.ts
blushi: sounds good, given this comment: https://github.com/regen-network/regen-web/pull/1551#discussion_r1021449051, that could possibly be addressed as part of regen-network/regen-registry#1400

web-registry/src/pages/Dashboard/MyEcocredits/hooks/useCreditSendSubmit.tsx
```diff
@@ -69,11 +83,32 @@ const useCreditSendSubmit = ({
         memo: note,
       };
 
-      const onError = (): void => {
-        track('sendFailure');
+      const onError = (err?: Error): void => {
+        const batchInfo = credits?.find(batch => batch.denom === batchDenom);
```
blushi, line 87: I believe you could just extract this outside `onError` and `onSuccess` and use `const batchInfo = credits[creditSendOpen];` so you don't have to use `find()` multiple times.

blushi (commented):
look good overall, could you resolve the conflicts?

wgwz:
@blushi took care of that latest conflicts here

flagrede (approved):
Nice work !

wgwz:
@Mergifyio backport redwood hambach master

PR #1798, opened 2023-02-21
feat: implement double-csrf pattern and enable withCredentials for the registry-server graphql client

wgwz:
**don't merge yet**: this PR is depedent on this registry-server PR being merged and deployed first:
https://github.com/regen-network/registry-server/pull/227

## Description

Closes: https://github.com/regen-network/regen-registry/issues/1556

This PR configures the graphql client to use the double-csrf pattern.
Additionally, it instructs the graphql client to include _any_ cookies from the registry server in graphql requests.

Important files to review are:
1. `web-registry/src/apollo.tsx`
    a. this file contains the logic to prepare the graphql client for making authenticated requests
2. `web-registry/src/lib/clients/apolloClientFactory.ts`
    a. this file contains the factory for creating an apollo client in our application
    b. this was added to solve a few problems
    c. it is a basis for the singleton pattern for the graphql client
3. `web-registry/src/index.tsx`
    a. there are some important adjustments related to configuring the graphl client
4. `web-registry/src/components/templates/ProjectDetails/ProjectDetails.loader.ts`
    a. this contains adjustments so that the singleton graphql client is utilized
    
This PR adds two ways of getting an authenticated graphql client:

Method 1: the `useApolloClient` hook in react function components ([example](https://github.com/regen-network/regen-web/blob/kyle/1535-add-csrf-to-graphql-client/web-registry/src/components/templates/ProjectDetails/ProjectDetails.tsx#L58))
Method 2: by calling the singleton `apolloClientFactory.getClient()` method in regular functions passed through props ([example](https://github.com/regen-network/regen-web/blob/kyle/1535-add-csrf-to-graphql-client/web-registry/src/components/templates/ProjectDetails/ProjectDetails.loader.ts#L41))

Related task:
- https://github.com/regen-network/regen-registry/issues/1535#issuecomment-1428651299

Testing instructions:
- Currently this can only be tested against a local version of registry-server running this code:
    - https://github.com/regen-network/registry-server/pull/290
- Without the changes that this PR introduces, graphql queries to the registry-server return an "invalid csrf token" error
- With the changes in this PR, graphql queries to the registry-server succeed with 200 status codes and expected responses

---

### Author Checklist

I have...

#### Setting up backport PRs

   your PR:

3. If your branch does have merge commits:

   ```
   git merge dev
   git merge dev
   ```

### How to test

1.

### Reviewers Checklist

I have...

- [ ] reviewed tests


flagrede (approved):
Looks good !

wgwz:
One thing that is not ideal about this setup is that there is a lot of chattiness between the app and the registry-server. I.e. the sequence below had 10 requests to get a CSRF token before issuing graphql call (and there are more not shown in the screenshot):

image

Only one CSRF request should be needed between the client app and the server.
The CSRF cookie and the CSRF token from that request can be cached for a while.

I tried a few things to add some caching to this but it didn't seem to be working:
- [caching lookups following the apollo docs](https://www.apollographql.com/docs/react/api/link/apollo-link-context/#caching-lookups)
- and i also tried using react-query but since this is a plain function it wasn't working

there's probably a much better way of accomplishing this.
and so for now maybe it's best to just leave this as something to extend once others pickup more work in the login system.

but i'd be glad to try work on this some more if others have suggestions for a path forward

/cc @flagrede @blushi

flagrede:
@wgwz Ow I didn't realize that this would trigger a call for each instance.
One way to solve this is to use a singleton pattern to reuse the same client instance.
Some documentation on that: https://www.patterns.dev/posts/singleton-pattern/

flagrede (commented):
I am changing my review state to comment while we correctly determine how to instantiate the client only once.

flagrede (commented):
Thanks for your work on this PR @wgwz this is way better already !
I just noticed that there are still 4 requests being done on the deploy preview instead of one. Any idea why?
image

wgwz:
> Thanks for your work on this PR @wgwz this is way better already ! I just noticed that there are still 4 requests being done on the deploy preview instead of one. Any idea why?

@flagrede thanks! i'm thinking that this might just be related to some implicit retry logic of react-query(?). i'll try to reproduce that locally and see what happens when i deploy a version of the server that doesn't have the /csrfToken endpoint yet. but a 404 is expected at least in this case for the deploy preview because staging does not have the /csrfToken endpoint yet.

wgwz:
> > Thanks for your work on this PR @wgwz this is way better already ! I just noticed that there are still 4 requests being done on the deploy preview instead of one. Any idea why?
> 
> i'm thinking that this might just be related to some implicit retry logic of react-query(?). i'll try to reproduce that locally and see what happens when i deploy a version of the server that doesn't have the /csrfToken endpoint yet. but a 404 is expected at least in this case for the deploy preview because staging does not have the /csrfToken endpoint yet.

@flagrede indeed, react-query will retry failed requests by default three times after the initial query failed:

https://react-query-v3.tanstack.com/guides/query-retries

flagrede:
@wgwz thanks for the update. Regarding `react-query` there should be no retry logic by default, it has to be explicitly specified. This is not a blocker but would be nice to understand why. If you can't figure out why, please create a separate issue for it so we can investigate it later on.

web-registry/src/apollo.tsx
```diff
@@ -1,84 +1,60 @@
...
-  });
-
-  const link = ApolloLink.from([withToken, authMiddleware.concat(httpLink)]);
-
-  const client = new ApolloClient({
+  const link = authLink.concat(httpLink);
+  apolloClientFactory.prepare({
     cache: new InMemoryCache(),
-    link,
+    link: authLink.concat(link),
```
blushi, line 53: why do we need another `.concat` here?
shouldn't this just be:
```suggestion
    link,
```

blushi (approved):
looking good, nice work! just a small comment

web-registry/src/apollo.tsx
wgwz: just a typo, good catch :-)

blushi:
@regen-network/interfaces-team can we merge this?

web-registry/src/apollo.tsx
```diff
@@ -1,84 +1,60 @@
...
-  });
-
-  const link = ApolloLink.from([withToken, authMiddleware.concat(httpLink)]);
-
-  const client = new ApolloClient({
+  const link = authLink.concat(httpLink);
+  apolloClientFactory.prepare({
     cache: new InMemoryCache(),
-    link,
+    link: link,
```
blushi, line 53:
```suggestion
    link,
```

web-registry/src/apollo.tsx
```diff
@@ -1,84 +1,60 @@
...
+    queryKey: ['csrfToken'],
+    queryFn: async () => {
+      const resp = await fetch(`${baseUri}/csrfToken`, {
+        method: 'GET',
+        credentials: 'include',
+      });
+      const { token } = await resp.json();
+      return { token, headers: resp.headers };
+    },
+  });
```
blushi, line 28: maybe we could extract this part into a reusable query hook?
anyway I can do that as part of https://github.com/regen-network/regen-web/pull/1814 too so we can merge this

wgwz: SGTM, let's do this there

blushi: done: https://github.com/regen-network/regen-web/pull/1814/commits/59e4f853776fbba9ab25d4c2a35e55cc2cf3782b

flagrede:
@blushi @wgwz Looks good to be merged.

blushi:
@wgwz will coordinate on merging this after the registry-server keplr login PR per his slack message: https://regen-network.slack.com/archives/C01LX9E8QN8/p1678735781413879

Issue #1848, opened 2023-04-04
Ecocredit batch displays in portfolio when no credits are held by user


clevinson:
**Describe the bug**

When testing out mainnet NCT transfers and basketing UX with Christian last week I ran into a UI bug where ecocredit projects were displayed on my portfolio when I did not have any credits.

You can view the portfolio live here (or see screenshot below):

https://app.regen.network/ecocredits/accounts/regen1l0ck43z80q7uzfd70qzsm9lggwvz3gdy774x30/portfolio

(**Side note**: you have to manually click on the "Portfolio" header section in order for the ecocredits to render at all, we should probably track this as a separate bug)

**To Reproduce**
I believe this can be replicated on testnet as well:

1. Start with some ecocredits from a batch that can be put into a basket
2. On portfolio page, select to "Put" all credits of that batch into the basket
3. See that the ecocredits table of all credits in your portfolio still displays a row for the batch with 0 credits in it, and in Amount Tradable, Amount Retired, Amount Escrowed, it displays 3 dashes "`-`" where the amounts would normally be

**Expected behavior**
After transferring all credits from a batch into a basket, the row corresponding to that ecocredit batch should disappear from the Ecocredits table.

**Screenshots**

<img width="1507" alt="Screenshot 2023-04-03 at 10 37 00 AM" src="https://user-images.githubusercontent.com/832188/229943900-ae62fc7e-6687-4aef-81bb-2c65c1a54e26.png">

**Desktop (please complete the following information):**
 - OS: 
 - Browser: 
 - Version: 

**Smartphone (please complete the following information):**
 - Device: 
 - OS: 
 - Browser: 
 - Version: 

**Additional context**

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels and zenhub epics applied
- [ ] Appropriate contributors tagged

blushi:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Regen-Registry-5f8998bec8958d000f4609e2/issues/regen-network/regen-web/1848?planning-poker) with Zenhub @flagrede

flagrede:
@blushi I'm guessing we might still want to display the line if the user has retired credits but 0 tradable? We might just want to hide the line in the case where he has 0 on all 3 columns, wdyt ?

PR #1853, opened 2023-04-05
fix: hide ecocredit batch when no balance

flagrede:
## Description

Closes: #1848

- hide ecocredits when the user has a balance of 0 of all 3 fields (tradable, retired, escrowed).

---

### Author Checklist

I have...

### How to test

1. go to https://deploy-preview-1853--regen-registry.netlify.app/ecocredits/accounts/regen1df675r9vnf7pdedn4sf26svdsem3ugavgxmy46/portfolio
2. Check that there's no line in the table without value on all 3 balance columns

### Reviewers Checklist

I have...

- [ ] reviewed tests

### DIFF
--- a/web-registry/src/pages/Dashboard/MyEcocredits/hooks/useFetchEcocredits.ts
+++ b/web-registry/src/pages/Dashboard/MyEcocredits/hooks/useFetchEcocredits.ts
@@ -133,15 +133,22 @@ export const useFetchEcocredits = ({
   // Normalization
   // isLoading -> undefined: return empty strings in normalizer to trigger skeleton
   // !isLoading -> null/result: return results with field value different from empty strings and stop displaying the skeletons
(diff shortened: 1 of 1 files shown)


blushi:
works fine, just one comment when testing this from my portfolio (with paginated ecocredits table) by putting all my ecocredits of a certain batch into a basket (with no retired or escrowed credits) then the table current page displays 4 lines instead of 5 so it feels a bit weird when navigating through the pages after that, but not sure if that's a real blocker

flagrede:
@blushi I noticed that too but unfortunately I don't think we can do much about it. The reason is that requests are done page by page and if one page contains a balance of 0/0/0 we can't replace it with something else unless we do another request but this start to feel over complicated.

blushi (approved):

PR #1867, opened 2023-04-12
perf: NextJS POC

flagrede:
## Description

Closes: regen-network/regen-registry#1595

- initialize Next project inside `web-www`
- migrate home page
- various improvements from the current page on responsive

# Performance improvement

| After+Before | Desktop (+28.75%) | Mobile (+55%) |
|--------------|---------|--------|
| Before       |  image | image |
| After       | image |  image |

[Link to full page speed report](https://pagespeed.web.dev/analysis/https-steady-dodol-4090c8-netlify-app/h4dorjst1h?form_factor=desktop)

### TBD in separate tasks
- there's still room for improvement on the image side, `next/image` is not yet used everywhere
- improve font loading
- define SEO correctly
- use src dir ? (most NextJS projects do not use src folder but we could if want to)

### Note

This POC does not use RSC as emotion is not fully ready for NextJS /app dir, but it could easily be migrated to it later on.
Using RSC should drastically reduce our bundle size and give us additional performance improvement.
Related issue: https://github.com/emotion-js/emotion/issues/2928

---

### Author Checklist

I have...

### How to test

1. https://steady-dodol-4090c8.netlify.app/

### Reviewers Checklist

I have...

- [ ] reviewed tests


web-components/src/components/tabs/IconTab.tsx
```diff
@@ -49,6 +49,7 @@ const IconTab = ({
   ...props
 }: IconTabProps): JSX.Element | null => {
   return hidden ? null : (
+    // @ts-ignore
```
flagrede, line 52: Had a new TS issue here, not sure why, might be because of a dependency upgrade. As it was not needed for the POC I did not investigate further for now.

web-www/pages/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+    }),
+    sanityClient.query<ValuesSectionQuery>({
+      query: ValuesSectionDocument,
+    }),
+    sanityClient.query<LedgerSectionQuery>({
+      query: LedgerSectionDocument,
+    }),
+    sanityClient.query<BlogSectionQuery>({
+      query: BlogSectionDocument,
+    }),
```
flagrede, line 120: In order to save some time for the POC, the same fetching queries from the current site have been reused.
We could improve this part by having a single query.

haveanicedavid (commented):
I see from the description that only the homepage was migrated, but I'm still having a hard time seeing what are next-specific changes, as it seems much of this diff seems like rearranging prior files. I'm sure we can go over this in an architecture call

The perf gain looks great though, and it doesn't seem like things would too complex to migrate, based on what i'm seeing here. Nice work!

flagrede:
@haveanicedavid I confirm that only the home page has been migrated so far. Since the PR is only new files we can't really see the diff with previous ones but the big part, besides figuring out a bunch of initial issues, was to migrate all Gatsby-specific code (fetching, images...).

Now that the architecture is set up and most of the issues have been solved it should be straightforward to migrate the rest indeed.

web-www/pages/index.tsx
flagrede: Actually, I already fix it, will be cleaner: https://github.com/regen-network/regen-web/pull/1867/commits/f5b37e8895ec8ca64d98c42e724648d549f735c7

web-components/src/components/tabs/IconTab.tsx
flagrede: I double check, IconTab component is still working correctly this is a pure type issue.

web-www/README.md
```diff
@@ -0,0 +1,38 @@
+This is a [Next.js](https://nextjs.org/) project bootstrapped with [`create-next-app`](https://github.com/vercel/next.js/tree/canary/packages/create-next-app).
```
blushi, line 1: should we update the main readme too?

web-www/components/templates/Layout/Layout.tsx
```diff
@@ -0,0 +1,47 @@
+/**
+ * Layout component that queries for data
+ * with Gatsby's useStaticQuery component
+ *
+ * See: https://www.gatsbyjs.org/docs/use-static-query/
+ */
+
```
blushi, line 7:
```suggestion
```

web-www/styles/font.css
```diff
@@ -0,0 +1,12 @@
+@import url('https://fonts.googleapis.com/css?family=Muli:800,900&display=swap');
```
blushi, line 1: so now we need those web-components css files twice so we can import them directly from _app?

blushi (commented):
looks good, thanks for initiating this!
just a UI issue in the footer, spacing are a bit different (up: current website, down: new version)
image

erikalogie:
> looks good, thanks for initiating this! just a UI issue in the footer, spacing are a bit different (up: current website, down: new version) image

I definitely prefer the original spacing

web-www/styles/font.css
flagrede: They're not imported twice, they used to be imported in the [RegenThemeProvider](https://github.com/regen-network/regen-web/blob/main/web-components/src/theme/RegenThemeProvider.tsx#L34) but we're not using this provider here as it was not SSR compatible.

flagrede:
@erikalogie @blushi reason I tried to adjust the spacing is that it was quite broken currently around MD breakpoint.
image

To fix that I tried to dynamically wrap the number of columns:
image

Since we're not happy with the spacing changes with this solution I revert to the previous one while switching to the one-column layout one breakpoint earlier to fix this issue instead.

erikalogie:
> @erikalogie @blushi reason I tried to adjust the spacing is that it was quite broken currently around MD breakpoint. image
> 
> To fix that I tried to dynamically wrap the number of columns: image
> 
> Since we're not happy with the spacing changes with this solution I revert to the previous one while switching to the one-column layout one breakpoint earlier to fix this issue instead.
> 
> image

Sounds good though I like how you solved the md breakpoint!

web-www/styles/font.css
flagrede: Yes sure, we'll no longer need the old provider once `web-registry` will be migrated to nextjs.

blushi:
The vertical spacing between the line items still seems a bit too small:
image

flagrede:
@blushi the line issue is now fixed.

blushi (approved):

Issue #1861, opened 2023-04-06
Fix CSRF failures related to apollo client initialization 


wgwz:
**Describe the bug**

When the application first loads, there are 403 "invalid csrf token" errors issued for `/graphql` requests.
This only happens when the `registry-server` has enabled CSRF protection for the `/graphql` endpoint.
This error does not always happen for all of the `/graphql` requests.
It usually only happens for the first few `/graphql` requests.

Even though some subsequent `/graphql` requests succeed,
the failure of these first few requests can result in missing information in the app.
For example, we first found this issue in production when viewing a public profile page.
It resulted in being unable to see the name and profile image for a specific users profile.

**To Reproduce**

1. In your local development environment make sure that the registry-server is running _**without**_ the `GRAPHIQL_ENABLED` environment variable set.
2. Navigate to a public profile page, i.e. `http://localhost:3000/ecocredits/accounts/<address>/portfolio`
3. Open the network requests tab, and view the requests to `/graphql` that failing.
4. In particular, note how the `x-csrf-token` header's value is `undefined`

**Expected behavior**

There is no reason that the `x-csrf-token` header should be `undefined` in any requests.
Our graphql client should never be initializing with that value.

**Screenshots**

![Screen Shot 2023-04-06 at 5 06 58 PM](https://user-images.githubusercontent.com/10120306/230494126-c1b8dcc4-04c6-4668-911f-45c4aad77ff9.jpg)

**Additional context**

The most relevant parts of the code include:
- `web-registry/src/apollo.tsx`
- `web-registry/src/lib/clients/apolloClientFactory.ts`
- `web-registry/src/index.tsx` (where `AuthApolloProvider` is initialized)

Another part of the code that may be helpful since it involves a request that requires a CSRF token:
- `web-registry/src/lib/wallet/hooks/useLogin.tsx`

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels and zenhub epics applied
- [ ] Appropriate contributors tagged

wgwz:
[...]

PR #1874, opened 2023-04-18
fix: issues with csrf graphql queries

wgwz:
## Description

Closes: https://github.com/regen-network/regen-web/issues/1861

Important comment: https://github.com/regen-network/regen-web/issues/1861#issuecomment-1513687326

---

### Author Checklist

I have...

### How to test

1. See https://github.com/regen-network/regen-web/issues/1861#issuecomment-1513687326

### Reviewers Checklist

I have...

- [ ] reviewed tests

### DIFF
--- a/web-registry/src/pages/EcocreditsByAccount/EcocreditsByAccount.tsx
+++ b/web-registry/src/pages/EcocreditsByAccount/EcocreditsByAccount.tsx
@@ -1,12 +1,6 @@
 import { useMemo } from 'react';
 import { Outlet, useLocation, useParams } from 'react-router-dom';
-import {
-  ApolloClient,
-  NormalizedCacheObject,
-  useApolloClient,
-} from '@apollo/client';
 import { Box, useTheme } from '@mui/material';
-import { useQuery } from '@tanstack/react-query';
 
 import { Flex } from 'web-components/lib/components/box';
 import BridgeIcon from 'web-components/lib/components/icons/BridgeIcon';
@@ -17,8 +11,8 @@ import { IconTabProps } from 'web-components/lib/components/tabs/IconTab';
 import { IconTabs } from 'web-components/lib/components/tabs/IconTabs';
 import { truncate } from 'web-components/lib/utils/truncate';
 
+import { usePartyByAddrQuery } from 'generated/graphql';
 import { getAccountUrl } from 'lib/block-explorer';
-import { getPartyByAddrQuery } from 'lib/queries/react-query/registry-server/graphql/getPartyByAddrQuery/getPartyByAddrQuery';
 
 import { getSocialsLinks } from 'pages/Dashboard/Dashboard.utils';
 import { usePartyInfos } from 'pages/ProfileEdit/hooks/usePartyInfos';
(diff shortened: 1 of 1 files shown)


haveanicedavid (approved):
utACK

blushi:
when trying out on the deploy preview, I still see the first graphql request which is failing because the x-csrf-token header is undefined:

image

web-registry/src/pages/EcocreditsByAccount/EcocreditsByAccount.tsx
```diff
@@ -36,16 +30,9 @@ export const EcocreditsByAccount = (): JSX.Element => {
...
-  const { data: partyByAddr } = useQuery(
-    getPartyByAddrQuery({
-      client: graphqlClient,
-      addr: accountAddress ?? '',
-      enabled: !!accountAddress && !!graphqlClient,
-    }),
-  );
+  const { data: partyByAddr } = usePartyByAddrQuery({
+    variables: { addr: accountAddress ?? '' },
+  });
```
flagrede, line 35: While this seems to work, I'm not comfortable relying on retries for ensuring that this call will eventually work. If for some reason we exceed the number of retries and the client still has no csrf token due to network latency this call would still fail.
We could check if we have a CSRF token before doing the request instead to ensure that the call would always work with various different network latencies.

flagrede:
> when trying out on the deploy preview, I still see the first graphql request which is failing because the x-csrf-token header is undefined:
> 
> image

We should not do this request in the first place without a csrf token and relies on retires to make it work. Checking CSRF token before doing the request would be safer to ensure this call will work.

wgwz:
> > when trying out on the deploy preview, I still see the first graphql request which is failing because the x-csrf-token header is undefined:
> > image
> 
> We should not do this request in the first place without a csrf token and relies on retires to make it work. Checking CSRF token before doing the request would be safer to ensure this call will always work.

I have tried and I don’t know how to get this to work. Switching to the generated usePartyByAddr seems to make so the profile page always works.

wgwz:
> when trying out on the deploy preview, I still see the first graphql request which is failing because the x-csrf-token header is undefined:
> 
> image

Yes, I have seen this too, but the public profile page is working with these changes. And it’s really only because of the switch to the generated usePartyByAddr

web-registry/src/pages/EcocreditsByAccount/EcocreditsByAccount.tsx
flagrede: Ok, I'll try to have a look to see If I can get the react-query version working otherwise we'll use that.

flagrede: @wgwz I was able to fix the issue with react-query, here's the PR: https://github.com/regen-network/regen-web/pull/1877
I think the problem you ran into was that there were several `partyByAddr`queries and they all need the CSRF safeguard to prevent putting invalid data in the cache.

wgwz:
Closing in favor of:

https://github.com/regen-network/regen-web/pull/1877

PR #1709, opened 2022-12-16
fix: 1475-1437: popover issue and bridge flag

flagrede:
## Description

Closes: regen-network/regen-registry#1437 regen-network/regen-registry#1475

These 2 fixes have been packed together because they were related code-wise.

- prevent the MenuHover component to move to the bottom right corner of the screen on the navigation change
- preload page chunk on link hover
- move UserMenu config to registry
- add feature flag for the bridge

---

### Author Checklist

_All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues._

I have...

- [ ] provided a link to the relevant issue or specification
- [ ] provided instructions on how to test
- [ ] reviewed "Files changed" and left comments if necessary
- [ ] confirmed all CI checks have passed
- [ ] once the PR is closed, set up backport PRs for `redwood`, `hambach` and optionally `master` (see below)

#### Setting up backport PRs

After merging your PR to `dev`, set up backports by doing the following:

1. If your branch does not have merge commits, add the following comment to
   your PR:

- If PR is a new feature: `@Mergifyio backport redwood hambach`.
- If PR is a bug fix: `@Mergifyio backport redwood hambach master`.

2. If your branch does have merge commits:

   a. Pull latest `dev`, `hambach` and `redwood` branches

   b. Create new branches for backports and merge `dev` (replace `<PR#>` with your PR #)

   ```
   git checkout -b hambach-backport-<PR#> hambach
   git merge dev
   git push origin hambach-backport-<PR#>
   git checkout -b redwood-backport-<PR#> redwood
   git merge dev
   git push origin redwood-backport-<PR#>`
   ```

   c. Open new PRs in regen-web targeting `hambach` and `redwood`, respectively.

   d. If PR is a bug fix repeat the same process for `master` branch.

### How to test

1. https://deploy-preview-1709--regen-registry.netlify.app/

### Reviewers Checklist

_All items are required. Please add a note if the item is not applicable and please **add
your handle next to the items reviewed if you only reviewed selected items**._

I have...

- [ ] confirmed all author checklist items have been addressed
- [ ] reviewed code correctness and readability
- [ ] verified React components follow DRY principles
- [ ] reviewed documentation is accurate
- [ ] reviewed tests
- [ ] manually tested (if applicable)

PR #1912, opened 2023-05-24
fix: menu dropdown that do not close properly

blushi:
## Description

Closes: regen-network/regen-registry#1689

- did some refactoring of some header related components (the previous structure/names were a bit hard to follow IMO)
- see this commit: ~~94d44cbbe3645e1ef90bb5817c5595e0b4b6fb20~~ https://github.com/regen-network/regen-web/pull/1912/commits/2a4d9d145964217c2308f74aecf862129ed6fdb7 for the actual fix

---

### Author Checklist

I have...

### How to test

https://deploy-preview-1912--regen-registry.netlify.app/

### Reviewers Checklist

I have...

- [ ] reviewed tests


wgwz:
@blushi could you explain the fix commit a bit? i'm curious to understand it better

web-components/src/components/header/components/HeaderMenuItem/HeaderMenuItem.Hover.tsx
```diff
@@ -67,43 +67,41 @@ const MenuHover = ({
         )}
         {renderTitle && renderTitle()}
       </span>
-      {popoverAnchor.current !== null && (
```
flagrede, line 70: I remember doing that to prevent the dropdown to appears in the bottom right screen corner but apparently it created a new issue. I could no longer reproduce the original issue on the deploy preview so I can guess it's safe to remove now.

flagrede (approved):
Nice refactor!
Regarding the fix not sure why the original issue describe [here](https://github.com/regen-network/regen-web/pull/1709) is gone but it's a good thing anyway.

One last thing, not directly concerning PR, I noticed that the dropdown menu is triggered only when hovering over the user icon and not the button. Would be nice to have the hover zone on the button if it's easy to do, otherwise, we can do that in a separate ticket.
image

web-components/src/components/header/components/HeaderMenuItem/HeaderMenuItem.Hover.tsx
blushi: oh actually just realized it does appear back again at the bottom right, but then disappears very quickly, which makes it hard to catch! so I need to look into this a bit closer...

blushi: so looking more closely at some mui example: https://mui.com/material-ui/react-popover/#mouse-over-interaction and  also react docs regarding documentation for using refs (in particular do not read while rendering, see "Pitfall" in https://react.dev/reference/react/useRef#usage), I switched to using state but still need to keep the ref to the menu item title so the popover is still displayed below it once we hover on the dropdown elements

PR #1918, opened 2023-06-01
feat: update project roles form to use profiles

blushi:
## Description

Closes: regen-network/regen-registry#1504
Closes: https://github.com/regen-network/regen-registry/issues/1604

---

### Author Checklist

I have...

### How to test

Login and go to https://deploy-preview-1918--regen-registry.netlify.app/ecocredits/projects to create a new project or edit one, until you get to the Roles form

### Reviewers Checklist

I have...

- [ ] reviewed tests


blushi:
@flagrede running into an issue with babel-loader using @uidotdev/usehooks but making this R4R because the main changes are already there

blushi:
@erikalogie see testing instructions above

web-registry/src/components/organisms/RolesForm/components/RoleField/RoleField.tsx
```diff
@@ -0,0 +1,188 @@
...
+        input: debouncedInputValue,
+      }),
+    );
+    const { data: partiesByAccountId } = useQuery(
+      getPartiesByAccountIdQuery({
+        client: graphqlClient,
+        id: accountId,
+        enabled: !!accountId && !!graphqlClient,
+      }),
+    );
```
flagrede, line 78: I'm wondering if it's fine to have components doing fetch directly. While this is convenient it prevents us to add stories for them. This component would have been interesting to visualize and test with mock data in storybook (if it's too much of a pain to refactor we can leave it).

flagrede (commented):
Great work, this was a complex one!
Code-wise PR looks good.
Still having some issues on the testing side.

flagrede (commented):
Great work, this was a complex one!
Code-wise PR looks good.
Still having some issues on the testing side.

blushi:
> @blushi I tried to add a new profile but got this error: `"message":"new row violates row-level security policy for table \"wallet\""`. This only happen when I specify a wallet address.

It seems like the db doesn't allow inserting wallet directly, this way it can only be done through `private.create_new_account` when logging in for the first time, but in this case, it's not really an actual logged in user. I'm wondering how we should handle that, thoughts @wgwz? The simplest way to handle that would be to add some RLS policy on wallet INSERT `WITH CHECK (true)`
Currently, I'm just creating a new `wallet` if some address is provided and creating the associated `party`, without any account linked to it. I guess we'd also want that, if at some point, some user tries to log in with this address, it would automatically link the account to the existing `party`/`wallet`.

> ~Also sometimes I'm seeing only a few test values in the new role select and some other times I'm seeing a lot more test values~ nevermind I got confused by the behavior here.

Yeah from the design, your profiles (from your current account) are displayed by default, maybe you got confused by this part?

> Seems like the profile I added does not appear there.

How did you add it? using "add profile"?

web-registry/src/components/organisms/RolesForm/components/RoleField/RoleField.tsx
```diff
@@ -0,0 +1,188 @@
...
+            }
+            getOptionLabel={option => (isProfile(option) ? option.name : '')}
+            renderOption={(props, option, state) => (
+              <li {...props} key={state.index}>
+                {isProfile(option)
+                  ? `${option.name} (${truncate(option.address)})`
+                  : option}
+              </li>
+            )}
+            filterOptions={x => x}
```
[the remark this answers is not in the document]
blushi: see https://mui.com/material-ui/react-autocomplete/#search-as-you-type

erikalogie:
> * /cc @erikalogie It might be a bit confusing to have the `all profiles` category not showing up by default. Maybe for this part to be less confusing we could display the category with one item saying `type to search in all profiles` ?

I think this is probably a good idea. WDYT @blushi?

I'm experiencing this bug where if I change the profile info, then leave the project edit screen and come back, I can no longer choose between all of my profiles in the dropdown:
https://github.com/regen-network/regen-web/assets/8460237/838517dd-66a1-4787-8802-bf7630bf8193

Also, if I try to simply add a wallet address in the field, i.e. "regen1fma4ef8yjzqjtq04n64x29l5kdga2x46jp88tz", which does have a profile not connected to my account (https://deploy-preview-1918--regen-registry.netlify.app/ecocredits/accounts/regen1fma4ef8yjzqjtq04n64x29l5kdga2x46jp88tz/portfolio) then it doesn't work.

I'm getting "something went wrong" when trying to add a profile using the popup, with and without the wallet address, which does not permit me to save.

blushi:
> I think this is probably a good idea. WDYT @blushi?

I'm not sure, because typing will also filter your own profiles.
I believe instead, we might want to make the placeholder text more informative, currently it's "Choose entity or search by name or address" in the design (which I forgot to update in the code yet btw...). Maybe something like "Choose from your profiles or search all profiles by name or address". This way the user gets what he needs to do without having to open the dropdown.

image

cc/ @flagrede 

> I'm experiencing this bug where if I change the profile info, then leave the project edit screen and come back, I can no longer choose between all of my profiles in the dropdown:
> https://github.com/regen-network/regen-web/assets/8460237/838517dd-66a1-4787-8802-bf7630bf8193

Looking into it

> Also, if I try to simply add a wallet address in the field, i.e. "regen1fma4ef8yjzqjtq04n64x29l5kdga2x46jp88tz", which does have a profile not connected to my account (https://deploy-preview-1918--regen-registry.netlify.app/ecocredits/accounts/regen1fma4ef8yjzqjtq04n64x29l5kdga2x46jp88tz/portfolio) then it doesn't work.
>

I could add it successfully but then the field value disappears after submission, I believe it's because the profile doesn't have any name, we should set it to "Unnamed" by default as well in this case.

> I'm getting "something went wrong" when trying to add a profile using the popup, with and without the wallet address, which does not permit me to save.

With the wallet address, needs this fix on registry-server: https://github.com/regen-network/registry-server/pull/346
Without the wallet address, I'm unable to reproduce it, could you give more details? (fields value, errors in the console or network tab) thanks!

blushi:
> * for the `add profile` it was from this action

@flagrede there was an issue with the pg function `get_parties_by_name_or_addr` to query that in case a party had no wallet_id, needed to use `LEFT JOIN` instead of a inner `JOIN`, fixed in https://github.com/regen-network/registry-server/pull/346/commits/238cf7ecc8b76009ef7e91ac8c4e0c8c1b6b299a, good catch!

web-registry/src/components/organisms/RolesForm/components/RoleField/RoleField.tsx
blushi: we can extract `getPartiesByAccountIdQuery` easily but with regards to `getPartiesByNameOrAddrQuery`, it seems a bit more complex, since it depends on `deboucedInputValue`, hence `inputValue` which is part of the component state, unless we set it as a component prop instead? and have some in between component like this:

```
const [inputProjectDeveloper, setInputProjectDeveloper] = useState('');
const debouncedInputProjectDeveloper = useDebounce(inputProjectDeveloper);
const { data: partiesProjectDeveloper } = useQuery(
  getPartiesByNameOrAddrQuery({
    client: graphqlClient,
    enabled: !!graphqlClient && !!debouncedInputProjectDeveloper,
    input: debouncedInputProjectDeveloper,
  }),
);

return <RoleField
  setInputValue={setInputValue}
  inputValue={inputValue}
  parties={partiesProjectDeveloper}
  {...props}
/>
```

erikalogie:
> 'm not sure, because typing will also filter your own profiles with the current implementation (see my suggestion just in the 2nd section below). I believe instead, we might want to make the placeholder text more informative, currently it's "Choose entity or search by name or address" in the design (which I forgot to update in the code btw...). Maybe something like "Choose from your profiles or search all profiles by name or address". This way the user gets what he needs to do without having to open the dropdown.
> 
> image
> 
> cc/ @flagrede

I think that this upgrade is good, but I think that was @flagrede likely had in mind was something like this: https://www.figma.com/file/TyhFaXqko9F8OvLEE4Zomg/Issuer-Create%2FEdit-Projects%2C-Credit-Class%2C-Batches?type=design&node-id=2525%3A61133&t=BrZnRkQjUhADyohN-1. So this way when you are typing you still the instructions for finding other profiles. I like this idea. 


> > I'm experiencing this bug where if I change the profile info, then leave the project edit screen and come back, I can no longer choose between all of my profiles in the dropdown:
> > https://github.com/regen-network/regen-web/assets/8460237/838517dd-66a1-4787-8802-bf7630bf8193
> 
> This is because there's some value so the profiles shown in the dropdown are filtered based on this value. If you want to see your profiles back again, you would have to remove the current value (using the cross icon). Should we always show your profiles instead no matter what's the input value?

Yeah I think it is more intuitive to be able to switch between the profiles easily once you click the dropdown arrow or put your cursor in the field. I would show the same dropdown as I linked above in this case, but of course with the name of the current profile in the field.





> I could add it successfully but then the field value disappears after submission, I believe it's because the profile doesn't have any name, we should set it to "Unnamed" by default as well in this case.

This seems to be working now.



> > I'm getting "something went wrong" when trying to add a profile using the popup, with and without the wallet address, which does not permit me to save.
> 
> With the wallet address, needs this fix on registry-server: [regen-network/registry-server#346](https://github.com/regen-network/registry-server/pull/346) Without the wallet address, I'm unable to reproduce it, could you give more details? (fields values, errors in the console or network tab) thanks!

Ok without the wallet address does seem to be working now, not sure what was happening before. However we do want a way to edit this entity with the "edit entity" link in this comp: https://www.figma.com/file/TyhFaXqko9F8OvLEE4Zomg/Issuer-Create%2FEdit-Projects%2C-Credit-Class%2C-Batches?type=design&node-id=2359%3A232240&t=BrZnRkQjUhADyohN-1.

blushi:
> Ok without the wallet address does seem to be working now, not sure what was happening before. However we do want a way to edit this entity with the "edit entity" link in this comp: https://www.figma.com/file/TyhFaXqko9F8OvLEE4Zomg/Issuer-Create%2FEdit-Projects%2C-Credit-Class%2C-Batches?type=design&node-id=2359%3A232240&t=BrZnRkQjUhADyohN-1.

Ok from the [latest design](https://www.figma.com/file/TyhFaXqko9F8OvLEE4Zomg/Issuer-Create%2FEdit-Projects%2C-Credit-Class%2C-Batches?type=design&node-id=2-4559&t=6Ajk6lnAkT6Tupfo-0), I had thought that we didn't want to allow that anymore.
@wgwz do you see any security issue coming from this requirement? Since the keplr login work, we're now have a security policy that only allows the account owner of a profile to edit it but in this case, there's no account, do you think it would be fine to add another security policy that would allow updates on parties without any account_id by auth_user's?
I believe the ability to update the wallet address could also involve some security concerns, because we don't want to allow any auth user to be able to update or delete an existing wallet address. In this case, I'm not sure what else we could verify to add a layer of security, unless we add some `creator_id` to `party` that would be the account id of the account that created it initially? Then we base our security policies checks on this...

In any case, probably best to tackle as part of a separate issue.

blushi:
> I think that this upgrade is good, but I think that was @flagrede likely had in mind was something like this: https://www.figma.com/file/TyhFaXqko9F8OvLEE4Zomg/Issuer-Create%2FEdit-Projects%2C-Credit-Class%2C-Batches?type=design&node-id=2525%3A61133&t=BrZnRkQjUhADyohN-1. So this way when you are typing you still the instructions for finding other profiles. I like this idea.

Yeah I got that but I just thought it would be a bit repetitive given the input placeholder and also confusing to display this in the dropdown results, anyway I can add that, just see my suggestion on figma.

erikalogie:
> > I think that this upgrade is good, but I think that was @flagrede likely had in mind was something like this: https://www.figma.com/file/TyhFaXqko9F8OvLEE4Zomg/Issuer-Create%2FEdit-Projects%2C-Credit-Class%2C-Batches?type=design&node-id=2525%3A61133&t=BrZnRkQjUhADyohN-1. So this way when you are typing you still the instructions for finding other profiles. I like this idea.
> 
> Yeah I got that but I just thought it would be a bit repetitive given the input placeholder and also confusing to display this in the dropdown results, anyway I can add that, just see my suggestion on figma.

Ok cool, yeah I liked your suggestion in Figma.

erikalogie:
> > Ok without the wallet address does seem to be working now, not sure what was happening before. However we do want a way to edit this entity with the "edit entity" link in this comp: https://www.figma.com/file/TyhFaXqko9F8OvLEE4Zomg/Issuer-Create%2FEdit-Projects%2C-Credit-Class%2C-Batches?type=design&node-id=2359%3A232240&t=BrZnRkQjUhADyohN-1.
> 
> Ok from the [latest design](https://www.figma.com/file/TyhFaXqko9F8OvLEE4Zomg/Issuer-Create%2FEdit-Projects%2C-Credit-Class%2C-Batches?type=design&node-id=2-4559&t=6Ajk6lnAkT6Tupfo-0), I had thought that we didn't want to allow that anymore. @wgwz do you see any security issue coming from this requirement? Since the keplr login work, we're now have a security policy that only allows the account owner of a profile to edit it but in this case, there's no account, do you think it would be fine to add another security policy that would allow updates on parties without any account_id by auth_user's? I believe the ability to update the wallet address could also involve some security concerns, because we don't want to allow any auth user to be able to update or delete an existing wallet address. In this case, I'm not sure what else we could verify to add a layer of security, unless we add some `creator_id` to `party` that would be the account id of the account that created it initially? Then we base our security policies checks on this...
> 
> In any case, probably best to tackle as part of a separate issue.

Hey sorry this was my fault, when I cleaned up the files I think I deleted it accidentally. But yeah, we can have this be part of a separate issue, no problem.

web-registry/src/components/organisms/RolesForm/components/RoleField/RoleField.tsx
blushi: ping @flagrede

web-registry/src/components/organisms/RolesForm/components/RoleField/RoleField.tsx
```diff
@@ -0,0 +1,188 @@
...
+    root?: string;
+  };
+  description?: string;
+  label?: string;
+  optional?: boolean | string;
+  setValue: (value: ProfileModalSchemaType | null) => void;
+  value?: ProfileModalSchemaType | null;
+}
+
+export const RoleField = forwardRef<HTMLInputElement, Props>(
```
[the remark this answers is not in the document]
flagrede: Not necessarily as props of `RoleField` but extract some props (the most lengthy ones) of `Autocomplete` into separate functions/components in external files.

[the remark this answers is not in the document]
blushi: yeah got it afterwards, thanks!

web-registry/src/components/organisms/RolesForm/components/RoleField/RoleField.tsx
flagrede: Sorry for the late answer, missed the notification.
I agree that the debounce value should be part of the component.
We could just pass the `setInputValue ` in your example and keep the debounce value in `RoleField`.
We would trigger the `setInputValue ` call in a `useEffect` in `RoleField` whenever the debounce value change.

flagrede:
> > I think that this upgrade is good, but I think that was @flagrede likely had in mind was something like this: https://www.figma.com/file/TyhFaXqko9F8OvLEE4Zomg/Issuer-Create%2FEdit-Projects%2C-Credit-Class%2C-Batches?type=design&node-id=2525%3A61133&t=BrZnRkQjUhADyohN-1. So this way when you are typing you still the instructions for finding other profiles. I like this idea.
> 
> Yeah I got that but I just thought it would be a bit repetitive given the input placeholder and also confusing to display this in the dropdown results, anyway I can add that, just see my suggestion on figma.

Whenever I tested I actually start typing before reading the placeholder, that's why I was confused after. I guess it can't hurt to be a bit repetitive with helpers' text for the user since we never know if they will use a component as we intend to.

web-registry/src/components/organisms/RolesForm/components/RoleField/RoleField.tsx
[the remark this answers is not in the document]
flagrede: Again if this is too much of a pain, you can ignore this suggestion.

web-registry/src/components/organisms/RolesForm/components/RoleField/RoleField.tsx
flagrede: Oh, I see the confusion, I meant that `inputValue` stays in `RoleField` too and the callback that would be passed to `RoleField` would be from another `useState` from the component using `RoleField` dedicated to `getPartiesByNameOrAddrQuery` query. Does that make sense ?
The Flow would be:
inputValue change -> debounce value change -> call callback from parent component with the latest debounce value

flagrede: This is just to save one prop from `RoleField` and separate the component state from the query param but your suggestion works too.

web-registry/src/components/organisms/RolesForm/components/RoleField/RoleField.tsx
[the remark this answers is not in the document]
blushi: done for most of the Autocomplete props

web-registry/src/components/organisms/RolesForm/components/RoleField/RoleField.tsx
blushi: oh I see I think that works

blushi: done, need to add a story for RoleField now but ready for another round of review
need the registry-server PR to be merged still before everything works as expected

blushi:
@erikalogie ready for another review

web-registry/src/components/organisms/RolesForm/components/RoleField/RoleField.tsx
```diff
@@ -0,0 +1,167 @@
...
+            getOptionLabel={getOptionLabel}
+            renderOption={(props, option) => (
+              <RoleFieldOption props={props} option={option} />
+            )}
+            filterOptions={x => x}
+            options={getOptions(options, setProfileAdd, accountId)}
+            groupBy={option => (isProfile(option) ? option.group : '')}
+            renderGroup={params => <RoleFieldGroup params={params} />}
+            autoComplete
+          />
```
flagrede, line 155: 👌

flagrede (commented):
Code looks great, nice job!
I have one last question when a value is selected in RoleField I can't delete it unless I use the clear button (the value is getting reset whenever I delete a character) is this the behavior we want?

blushi:
@erikalogie @flagrede this should be fixed now

blushi:
@erikalogie could you have a final look at this before merging? thanks

PR #1919, opened 2023-06-05
feat: add certificate table and page

flagrede:
## Description

Closes: regen-network/regen-registry#425

add:
- Certificate table on My Portfolio
- Certificate page with printable option

---

### Author Checklist

I have...

### How to test

1. table: https://deploy-preview-1919--regen-registry.netlify.app/ecocredits/portfolio
2. page: https://deploy-preview-1919--regen-registry.netlify.app/certificate/C01-001 

### Reviewers Checklist

I have...

- [ ] reviewed tests


web-registry/src/pages/Certificate/Certificate.tsx
```diff
@@ -0,0 +1,69 @@
...
+          </Title>
+          <ShareIcons
+            xsSize={theme.spacing(10)}
+            url={`${window.location.origin}/buyers`}
+          />
+        </Grid>
+        <Grid item xs={12} sm={6}>
+          <OutlinedButton
+            className={classes.printButton}
+            onClick={() => window.print()}
```
flagrede, line 59: I'm using `print` from the browser here instead of `react-to-print` as it was a real struggle to format the layout with the library while using the browser native format worked better.

erikalogie:
> @erikalogie you can check the UI for:
> 
> * the retirement certificates table here: https://deploy-preview-1919--regen-registry.netlify.app/ecocredits/portfolio
> * the certificate page here: https://deploy-preview-1919--regen-registry.netlify.app/certificate/C01-001 (you can check the print mode as well).

Looking really good!!

@flagrede the printable certificate is looking pretty funky: https://monosnap.com/file/KPNYOquDn2sdmvHznqRnXUnzXdijV1. Is there a way to have it appear on just one page?

The tooltip for the "ecocredit batch ID" should be "Each credit batch has a unique ID (i.e. denomination) that starts with the abbreviation of the credit type followed by the project ID, start date and end date of the vintage year, and ending in the batch sequence number. " It would be nice if the header were to match on the "Ecocredits" table, so "Ecocredit batch ID" instead of "batch denom", with the tooltip. This can be a separate task though.

On the retirement certificate, I guess linking the "tx hash", "project", "credit class", and "retired by" user profile to their respective pages will be done after #1695 is complete?

On the mobile version of the retirement certificate page, could we have the action button just be "view" instead of "view certificate" so it is easier to see more of the table?

flagrede:
@erikalogie landscape mode is now better! It was hard to get everything to fit into the page so I had to remove some margins and reduce font sizes, let me know if that looks ok.

erikalogie:
> @erikalogie landscape mode is now better! It was hard to get everything to fit into the page so I had to remove some margins and reduce font sizes, let me know if that looks ok.

Looks great to me!

web-registry/src/components/organisms/Portfolio/Portfolio.tsx
```diff
@@ -0,0 +1,111 @@
...
+                  navigate(`/certificate/${credits?.[index].projectId}`)
+                }
+              />
+            )}
+            onTableChange={onTableChange}
+            initialPaginationParams={initialPaginationParams}
+            isIgnoreOffset={isIgnoreOffset}
+          />
+        ),
+        hidden: location.pathname !== '/ecocredits/portfolio',
```
flagrede, line 83: will hide the tab everywhere before merging while we get real data.

web-registry/src/components/organisms/EcocreditsTable.tsx
```diff
@@ -62,14 +54,19 @@ export const EcocreditsTable: React.FC<
...
+            alignItems: 'center',
             minWidth: {
               xs: 'auto',
               sm: '11rem',
               lg: '13rem',
             },
           }}
         >
-          Batch Denom
+          <BreakText sx={{ maxWidth: 76 }}>{'Ecocredit batch ID'}</BreakText>
```
blushi, line 66: Is this intended to turn "batch denom" into "ecocredit batch ID"? why not using the term used on regen-ledger? has there been any user feedback on it not being clear enough?
If we want to change this, this should be done everywhere so we don't end up with inconsistencies which could confuse users even more:
- credit batch page
- buy/sell forms
- basket table
- cancel
- create batch
...

cc/ @erikalogie

blushi (approved):
Pre-approving from a code standpoint, looks great! but I think we should do something about the renaming of batch denom to ecocredit batch ID before merging

web-registry/src/components/organisms/EcocreditsTable.tsx
flagrede: @blushi I can definitely revert the change to the ecocredit table while we have this discussion. The new retirement table won't be visible too for now so this new naming would not be visible anywhere.

blushi: sounds good, then we can do the larger renaming in another PR

Issue #1963, opened 2023-07-10
Implement remaining queries based on indexer for improved load performance


ryanchristo:
- Txs query on stats page - activity table
- Adding tx hash to batch data (`addDataToBatches`)

ryanchristo:
@blushi can you identify the queries we need to update to complete epic #977? :pray:

blushi:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RND-Dev-Team-5f8998bec8958d000f4609e2/issues/regen-network/regen-web/1963?planning-poker) with Zenhub @flagrede @ryanchristo @wgwz

ryanchristo:
Work started in #2018 but dependent upon https://github.com/regen-network/indexer/issues/32.

PR #2018, opened 2023-08-09
feat: use indexer for activity page

flagrede:
## Description

Closes: regen-network/regen-web#1963

Implement `allTxes` query from `indexer` endpoint to enhance the loading performance of the activity table.

---

### Author Checklist

I have...

### How to test

1. https://deploy-preview-2018--regen-marketplace.netlify.app/stats/activity

### Reviewers Checklist

I have...

- [ ] reviewed tests


flagrede:
@erikalogie You can check that the activity table is loading faster now: https://deploy-preview-2018--regen-registry.netlify.app/stats/activity
The rows that have no message are mostly failed transactions and have no logs so we can't get the message from them, what should we do about them?

cc @blushi @ryanchristo

flagrede:
@blushi  @ryanchristo  Regarding the second part of the issue: Adding tx hash to batch data (addDataToBatches)
Unless I missed something I think we unfortunately can't use the indexer for that because we could need to get all txes with the `data` attribute (which is big since this is a stringify json) which creates a payload too big for the server to handle.

flagrede:
@erikalogie I will check that.

erikalogie:
Second option is we don't worry about it for now and open a separate ticket

flagrede:
@erikalogie Ok so I tried that and we end up with a lot of blank pages. Since this is paginated we can't know ahead of time if the page will be empty or not. Also by filtering the failed transactions the total count becomes incorrect and we can't fix that easily. I guess we'll have to keep them for now, maybe we could display a "failed transaction" in the message column as a default value if there's no logs?

blushi:
> > The rows that have no message are mostly failed transactions and have no logs so we can't get the message from them, what should we do about them?
> 
> I'm not sure, can we filter them out easily?

We should filter the txs, I don't think we want to display all the possible txs but just the ones which have msgs relevant to the ecocredit module, just like it was the case before.

blushi:
Another option could also be to apply a filter/graphql condition on the `tx.data` jsonb column. That doesn't seem to be enabled by default by postgraphile but this is something that is possible to add: https://www.graphile.org/postgraphile/filtering/
We would also need to evaluate how performant it is.

flagrede:
@blushi For the first solution we would use `allMsgEvents` instead of `allTxes` and call `txByChainNumAndBlockHeightAndTxIdx` for each row (about ~11 queries by page with default settings). Tx can have multiple  "message" events so it might be tricky to filter with pagination as well.

The second option would be ideal given that it has decent performance.

flagrede:
@blushi here's an example query to get all action events. Would it be possible to keep only the first entry for each block heigh to make it work with pagination?

```graphql
{
  allMsgEventAttrs (first: 5, condition: {key: "action"}, orderBy: BLOCK_HEIGHT_DESC,) {
    nodes {
      type
      key
      value
      blockHeight
      txIdx
    }
  }
}
```

blushi:
> @blushi here's an example query to get all action events. Would it be possible to keep only the first entry for each block heigh to make it work with pagination?
> 
> ```graphql
> {
>   allMsgEventAttrs (first: 5, condition: {key: "action"}, orderBy: BLOCK_HEIGHT_DESC,) {
>     nodes {
>       type
>       key
>       value
>       blockHeight
>       txIdx
>     }
>   }
> }
> ```

Why would we only want to keep the first entry?

flagrede:
@blushi By keeping the first entry only (the one we use to get the message name) we would be able to paginate.

ryanchristo:
Noting we've decided to use the graphql filter but it's dependent upon https://github.com/regen-network/indexer/issues/32.

ryanchristo:
Also looking good. Can you rebase in the meantime so it's easier to test locally?

ryanchristo:
This should be unblocked now that #34 is complete.

wgwz:
Hey @flagrede just wanted to add that this query is working quite nicely, and it's available in regen-server staging environment (https://api-staging.regen.network/indexer/v1/graphiql) 

Query for the first page:
```graphql
{
  allEcocreditTxes(first: 10) {
    nodes {
      blockHeight
      hash
    }
    pageInfo {
      endCursor
      hasNextPage
    }
  }
}
```

Query for the second page using the endCursor from the first query:
```graphql
{
  allEcocreditTxes(first: 10, after: "WyJuYXR1cmFsIiwxMF0=") {
    nodes {
      blockHeight
      hash
    }
    pageInfo {
      endCursor
      hasNextPage
    }
  }
}
```

Note: performance seems quite good for small pagination sizes.
Note: if you add `totalCount` parameter to the query, it affects performance negatively.

Hope this works well! Let me know if there's any questions or issues that come up.

wgwz:
Hey @flagrede this is actually not working quite as it's need to yet. Please consider this as still blocked until further notice

wgwz:
@flagrede this is now unblocked by https://github.com/regen-network/indexer/pull/35. everything i said earlier about those graphql queries is still accurate. the only thing to take note of is that the query always returns the most recent blocks first now (most recent activity should come first)

flagrede:
@wgwz thanks for the info, I have just updated the PR with the new query and it works well. I'm just waiting for the PR with the index change to BTREE to put this one back in QA.

flagrede:
@erikalogie the performance improvement for the activity page is ready for QA: https://deploy-preview-2018--regen-marketplace.netlify.app/stats/activity

web-marketplace/src/components/organisms/CreditActivityTable/CreditActivityTable.tsx
```diff
@@ -25,9 +31,11 @@ import { Label } from 'web-components/lib/components/typography';
 import { formatNumber } from 'web-components/lib/utils/format';
 import { truncate } from 'web-components/lib/utils/truncate';
 
-import { getHashUrl } from '../../lib/block-explorer';
-import { getEcocreditTxs, getReadableMessages } from '../../lib/ecocredit/api';
-import { ledgerRESTUri } from '../../lib/ledger';
+import { getAllEcocreditTxesQuery } from 'lib/queries/react-query/registry-server/graphql/indexer/getAllEcocreditTxes/getAllEcocreditTxes';
+
+import { getHashUrl } from '../../../lib/block-explorer';
+import { ledgerRESTUri } from '../../../lib/ledger';
```
ryanchristo, line 37:
```suggestion
import { getHashUrl } from 'lib/block-explorer';
import { ledgerRESTUri } from 'lib/ledger';
```

web-marketplace/src/lib/queries/react-query/registry-server/graphql/indexer/getAllEcocreditTxes/getAllEcocreditTxes.ts
```diff
@@ -0,0 +1,39 @@
...
+        variables: {
+          first,
+          offset,
+        },
+      });
+
+      return data;
+    } catch (e) {
+      return null;
+    }
```
ryanchristo, line 36: Should we be alerting sentry or logging the error?

web-marketplace/src/lib/queries/react-query/registry-server/graphql/indexer/getAllTxes/getAllTxes.ts
```diff
@@ -0,0 +1,42 @@
...
+          first,
+          offset,
+          orderBy,
+        },
+      });
+
+      return data;
+    } catch (e) {
+      return null;
+    }
```
ryanchristo, line 39: Same as above, should we be monitoring with sentry or logging the error?

web-marketplace/src/graphql/indexer/AllTxes.graphql
```diff
@@ -0,0 +1,11 @@
+query IndexerAllTxes($first: Int, $offset: Int, $orderBy: [TxesOrderBy!]) {
```
ryanchristo, line 1: What are we using this for? It looks like we never use `getAllTxes`. Any reason for adding this now?

ryanchristo (commented):
Looking good. A couple questions, the main one being the implementation of `getAllTxes`, which appears to have been added but is currently not being used.

erikalogie:
This looks great!

web-marketplace/src/graphql/indexer/AllTxes.graphql
flagrede: This is the initial query provided by the indexer before we added `allEcocreditTxes`. After switching to `allEcocreditTxes` I kept the code for `allTxes` since we might still need it later on.

wgwz (approved):
LGTM!

web-marketplace/src/graphql/indexer/AllTxes.graphql
ryanchristo: Ok, sounds good. Just wanted to make sure I wasn't missing something.

ryanchristo (approved):
tACK

PR #2042, opened 2023-08-24
chore: update local dev environment

ryanchristo:
## Description

This pull request includes fixes and improvements to the local dev environment:

- [x] adds missing basket bank denom metadata
- [x] database initialization only happens on build and first start
- [x] resolves auth user permissions not working
- [x] resolves block 0 indexer errors on startup
- [x] seeds indexer with class issuers from genesis state

---

### Author Checklist

I have...

### How to test

1. Create a dump file for restoring the production db

```
pg_dump [DATABASE_URL] --file dump-server-production -F c
```

2. build and start the local dev environment

```
docker-compose up
```

3. start the marketplace application

```
yarn start
```

4. Go through project creation flow (keep in mind #2047)

### Reviewers Checklist

I have...

- [ ] reviewed tests


docker/scripts/server_init.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# run migrations
+yarn migrate
+
+# TODO: investigate and resolve issue with local permission setup and then remove the following
+psql postgres://postgres:password@localhost:5432/server -c 'GRANT ALL PRIVILEGES ON TABLE credit_class TO auth_user;'
+psql postgres://postgres:password@localhost:5432/server -c 'GRANT ALL PRIVILEGES ON TABLE credit_class_version TO auth_user;'
+psql postgres://postgres:password@localhost:5432/server -c 'GRANT ALL PRIVILEGES ON TABLE project TO auth_user;'
+psql postgres://postgres:password@localhost:5432/server -c 'GRANT ALL PRIVILEGES ON TABLE organization TO auth_user;'
```
ryanchristo, line 10: This is a workaround and not the correct privileges. Maybe this is ok for now but not sure why privileges are not correct following the database restore and migrations.

Without this workaround, I am unable to create a project using the local dev environment:


Screenshot from 2023-08-28 11-51-00

Screenshot from 2023-08-28 11-50-20

ryanchristo: @wgwz or @blushi any ideas as to what the initialization issue might be here?

ryanchristo:
Other than the authorization workaround, I think this is ready for review. The main goal of these updates were to get the project creation flow working locally to test and debug some of the issues that came up last week. I made some other improvements as well in the process, including the use of initialization scripts to prevent running migrations on an existing container.

docker/scripts/server_init.sh
wgwz: what's the command you're using to initialize the database from scratch?
see here:
https://github.com/regen-network/regen-server/blob/dev/migrations/README.md#local-development

> 
> Then, you must initialize the database:
> ```
> $ export DATABASE_URL="postgres://postgres:postgres@localhost:5432/regen_registry"
> $ export SHADOW_DATABASE_URL="postgres://postgres:postgres@localhost:5432/regen_registry_shadow"
> $ export ROOT_DATABASE_URL="postgres://postgres:postgres@localhost:5432/postgres"
> $ yarn run graphile-migrate reset --erase
> ```

whereas, any time after that initial setup you need to use `yarn run graphile-migrate migrate` instead (or `yarn run graphile-migrate watch` if you want real-time application of new migrations)

you can see the same lines you added to get it working in the `initial_schema.sql` file:
https://github.com/regen-network/regen-server/blob/dev/migrations/initial_schema.sql#L1175

so it does seem like you might just be running the wrong command?

wgwz: i see the command now, so yeah, if you switch `yarn migrate` to `yarn run graphile-migrate reset --erase` you should be good

wgwz: here's how we set up for CI in the regen-server repo:
https://github.com/regen-network/regen-server/blob/dev/.github/workflows/registryServer.yml#L70

ryanchristo: Thanks! We do want to run `migrate` at this stage (the first time we start the server container) and avoid resetting the database. This step follows the restoring of the production database (which happens in `db_init.sh`), so this would be similar to us running migrations at the time of promoting a new commit to the production deployment.

I can try to make the setup a bit more clear between init files. Still investigating though.

wgwz: ok, well at some point, in order for the database to be working 100% correct that command does need to get run (otherwise the database would be missing stuff like the privileges). this is just sort of how it works with graphile-migrate. that command only needs to be run once, the first time you are preparing your local database or new database. and then for all subsequent changes after that it's `yarn migrate`.

ryanchristo: Thanks! This resolved the issue. I had to restructure things a bit to run `pg_restore` within the server container after running `graphile-migrate reset --erase` but that did the trick.

ryanchristo:
Ok, this is now ready to review again with workarounds resolved.

wgwz (approved):
utACK, sorry i didn't catch the issue with the graphile-migrate command sooner!

Issue #2039, opened 2023-08-22
A user should be able to add/update a project created outside the app


ryanchristo:
**Is your feature request related to a problem? Please describe.**

A user should be able to add and/or update a project created outside the application. For example, a project created using the CLI should then be available to update in the application if the logged in user is the project admin.

**Describe the solution you'd like**

One approach would be to have the project appear in the list of projects using the on-chain project id and then upon selecting the project the off-chain project id and database entry is created and the user continues through the edit project workflow.

**Describe alternatives you've considered (optional)**

TBD

**Additional context (optional)**

This recently occurred with the first Fibershed project on the testnet and resulted in having to manually add database entries in order to edit the project within the application.

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels and zenhub epics applied
- [ ] Appropriate contributors tagged

blushi:
Side note regarding the implementation: when updating values that are stored in the project unanchored metadata from the project edit form, we should create the project off-chain if not already there (and not only update the metadata)

ryanchristo:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RND-Dev-Team-5f8998bec8958d000f4609e2/issues/regen-network/regen-web/2039?planning-poker) with Zenhub @blushi @flagrede @ryanchristo @wgwz

PR #2060, opened 2023-08-31
fix: A user should be able to add/update a project created outside the app 

blushi:
## Description

Closes: #2039

---

### Author Checklist

I have...

### How to test

On https://deploy-preview-2060--regen-marketplace.netlify.app/ecocredits/projects, all the projects that you are an admin of should appear (before the list was based on offchain projects).

For devs:
- create a new project on chain from the CLI (with ou without metadata, both are supported now too)
- try to edit this project from the UI

For @erikalogie 
You can try to edit this project which doesn't have any corresponding off-chain project yet nor any anchored metadata available on staging: https://deploy-preview-2060--regen-marketplace.netlify.app/project-pages/C01-004/edit
But be aware that as soon as you edit some info, it will create the associated off-chain project and/or update the on-chain metadata.
Verify the edited info shows up in the project edit forms initial values and on the project page.

### Reviewers Checklist

I have...

- [ ] reviewed tests


blushi:
@erikalogie see testing instructions above

web-marketplace/graphql.schema.json
wgwz, line 1: i was wondering what the new `Myproject` object introduced here is, it looks like it might have been an experiment?

web-marketplace/src/generated/graphql.tsx
```diff
@@ -3165,6 +3325,8 @@ export enum ProjectsOrderBy {
   AdminWalletIdDesc = 'ADMIN_WALLET_ID_DESC',
   VerifierIdAsc = 'VERIFIER_ID_ASC',
   VerifierIdDesc = 'VERIFIER_ID_DESC',
+  TestAsc = 'TEST_ASC',
+  TestDesc = 'TEST_DESC',
```
ryanchristo, line 3329: This also looks suspicious.

ryanchristo (commented):
tACK. Pending generated code cleanup. Otherwise looks good to me.


Screenshot from 2023-09-05 13-51-27

Screenshot from 2023-09-05 13-52-37

web-marketplace/graphql.schema.json
blushi: yeah good catch, some custom view I had created for testing some time ago

web-marketplace/src/generated/graphql.tsx
blushi: fixed!

ryanchristo (approved):

blushi:
I've actually spotted a server related issue with this too.
When trying to edit an on-chain project (created through the CLI, without any corresponding off-chain project), if the user tries to add some image within the media form first, the /files endpoint will return an unauthorized error. This is because before we upload the file to s3, we verify that the current user is the admin of the project through db relationships: https://github.com/regen-network/regen-server/blob/424d6c59055573254cc4369317d982b640804f1e/server/routes/files.ts#L79. But the off-chain project will only get created on submit.
This is quite of an edge case but this should be addressed nevertheless.

Maybe a relatively simple solution would be to create the off-chain project before we try to upload using /files endpoint, so I'll add that as part of this PR.

blushi:
> I've actually spotted a server related issue with this too. When trying to edit an on-chain project (created through the CLI, without any corresponding off-chain project), if the user tries to add some image within the media form first, the /files endpoint will return an unauthorized error. This is because before we upload the file to s3, we verify that the current user is the admin of the project through db relationships: https://github.com/regen-network/regen-server/blob/424d6c59055573254cc4369317d982b640804f1e/server/routes/files.ts#L79. But the off-chain project will only get created on submit. This is quite of an edge case but this should be addressed nevertheless.
> 
> Maybe a relatively simple solution would be to create the off-chain project before we try to upload using /files endpoint, so I'll add that as part of this PR.

This should be fixed now.
I've also fixed some cache invalidation when editing the project anchored metadata and then coming back to MyProjects, how to test this is described here: https://github.com/regen-network/regen-web/pull/2078#discussion_r1317436639

wgwz (approved):
utACK

Issue #2084, opened 2023-09-07
Anchored project metadata gets overwritten by unanchored project page metadata on create and edit project


S4mmyb:
**Describe the bug**
When I used the project create form to create KSH01-001, I input the JSON-LD for the anchored metadata. When the project was created I noticed that none of it was populating in the additional info section. When I tried to edit the metadata post-project creation using the project edit form, I noticed that most of the metadata I submitted did not populate, and instead the JSON-LD anchored on-chain was a mix of the project page metadata and the anchored project metadata. 

This is what I originally submitted: 
```
{
  "@context": {
    "schema": "http://schema.org/",
    "regen": "https://schema.regen.network#",
    "qudt": "http://qudt.org/schema/qudt/",
    "unit": "http://qudt.org/vocab/unit/",
    "xsd": "http://www.w3.org/2001/XMLSchema#",
    "schema:url": {
      "@type": "schema:URL"
    },
    "qudt:unit": {
      "@type": "qudt:Unit"
    },
    "qudt:numericValue": {
      "@type": "xsd:double"
    },
    "regen:projectStartDate": {
      "@type": "xsd:date"
    },
    "regen:projectEndDate": {
      "@type": "xsd:date"
    },
    "regen:monitoringFrequency": {
      "@type": "schema:Duration"
    }
  },
  "@type": "regen:KSH01-Project",
  "schema:name": "Grgich Hills Estate Regenerative Sheep Grazing",
  "schema:location": {
    "@context": {
      "type": "@type",
      "@vocab": "https://purl.org/geojson/vocab#",
      "coordinates": {
        "@container": "@list"
      },
      "bbox": {
        "@container": "@list"
      }
    },
    "type": "Feature",
    "place_name": "Rutherford, California",
    "geometry": {
      "type": "Point",
      "coordinates": [
          -122.3754,
          38.4082
      ]
    }
  },
  "regen:projectSize": {
    "qudt:unit": "unit:HA",
    "qudt:numericValue": 126.6
  },
  "regen:projectActivity": "High Density, Short Duration Sheep Grazing",
  "regen:projectType": "Environmental Stewardship",
  "regen:projectStartDate": "2019-01-01",
  "regen:projectEndDate": "2023-05-31"
}
```

This is what was anchored: 
```
{
  "@type":"regen:KSH01-Project",
  "@context":{
    "xsd":"http://www.w3.org/2001/XMLSchema#",
    "qudt":"http://qudt.org/schema/qudt/",
[...]

PR #2110, opened 2023-09-18
fix: update project metadata submit to account for @context overwrite

blushi:
## Description

Closes: #2084

The issue described in #2084 can happen as soon as you completely overwrite the metadata as part of the metadata form, meaning that the context was getting overwritten too, which causes the part of the `@context` related to the unanchored metadata to be lost, causing issues with some unanchored metadata fields which require some `@context` (eg for defining a field as a `@list`).
This is because the current metadata form wasn't meant for overwriting the metadata completely, eg providing fields such as `schema:name` that already get filled up as part of other forms, but it was meant primarily for adding custom fields that cannot be provided with other forms. Now it should just supports both use cases.

---

### Author Checklist

I have...

### How to test

Create a new project from https://deploy-preview-2110--regen-marketplace.netlify.app/profile/projects
Go through the various steps, using "Rutherford, California, United States" as location and the following metadata in the final Metadata form (this is more or less what @S4mmyb had tried to submit which caused the issue):
```json
{
  "@context": {
    "schema": "http://schema.org/",
    "regen": "https://schema.regen.network#",
    "qudt": "http://qudt.org/schema/qudt/",
    "unit": "http://qudt.org/vocab/unit/",
    "xsd": "http://www.w3.org/2001/XMLSchema#",
    "schema:url": {
      "@type": "schema:URL"
    },
    "qudt:unit": {
      "@type": "qudt:Unit"
    },
    "qudt:numericValue": {
      "@type": "xsd:double"
    },
    "regen:projectStartDate": {
      "@type": "xsd:date"
    },
    "regen:projectEndDate": {
      "@type": "xsd:date"
    },
    "regen:monitoringFrequency": {
      "@type": "schema:Duration"
    }
  },
  "@type": "regen:KSH01-Project",
  "schema:name": "Grgich Hills Estate Regenerative Sheep Grazing",
  "schema:location": {
    "@context": {
      "type": "@type",
      "@vocab": "https://purl.org/geojson/vocab#",
      "coordinates": {
        "@container": "@list"
      },
      "bbox": {
        "@container": "@list"
      }
    },
    "type": "Feature",
    "place_name": "Rutherford, California",
    "geometry": {
      "type": "Point",
      "coordinates": [
          -122.3754,
          38.4082
      ]
    }
  },
  "regen:projectSize": {
    "qudt:unit": "unit:HA",
    "qudt:numericValue": 126.6
  },
  "regen:projectActivity": "High Density, Short Duration Sheep Grazing",
  "regen:projectType": "Environmental Stewardship",
  "regen:projectStartDate": "2019-01-01",
  "regen:projectEndDate": "2023-05-31"
}
```
Create the project on-chain and go check the project page that the additional info does appear based on the metadata above.

### Reviewers Checklist

I have...

- [ ] reviewed tests

### DIFF
--- a/web-marketplace/src/pages/ProjectReview/ProjectReview.tsx
+++ b/web-marketplace/src/pages/ProjectReview/ProjectReview.tsx
@@ -5,6 +5,7 @@ import { useApolloClient } from '@apollo/client';
 import { DeliverTxResponse } from '@cosmjs/stargate';
 import { Box, Card, CardMedia, useMediaQuery, useTheme } from '@mui/material';
 import { useQuery, useQueryClient } from '@tanstack/react-query';
+import { omit } from 'lodash';
 
 import ErrorBanner from 'web-components/lib/components/banner/ErrorBanner';
 import { ReviewCard } from 'web-components/lib/components/cards/ReviewCard/ReviewCard';
@@ -24,6 +25,7 @@ import {
   qudtUnit,
 } from 'lib/rdf';
 
+import { OMITTED_METADATA_KEYS } from 'pages/ProjectMetadata/ProjectMetadata.config';
 import {
   STORY_LABEL,
   STORY_TITLE_LABEL,
@@ -283,7 +285,10 @@ export const ProjectReview: React.FC<React.PropsWithChildren<unknown>> = () => {
               overflowY: 'scroll',
             })}
           >
-            <pre>{!!metadata && JSON.stringify(metadata, null, 2)}</pre>
+            <pre>
+              {!!metadata &&
+                JSON.stringify(omit(metadata, OMITTED_METADATA_KEYS), null, 2)}
+            </pre>
           </Box>
         )}
       </ReviewCard>
--- a/web-marketplace/src/pages/ProjectMetadata/hooks/useProjectMetadataSubmit.tsx
+++ b/web-marketplace/src/pages/ProjectMetadata/hooks/useProjectMetadataSubmit.tsx
@@ -31,8 +31,12 @@ export const useProjectMetadataSubmit = ({
   const projectMetadataSubmit = useCallback(
     async ({ values }: SubmitParams): Promise<void> => {
       const parsedMetaData = JSON.parse(values.metadata);
-      const baseMetadata = pick(metadata, OMITTED_METADATA_KEYS);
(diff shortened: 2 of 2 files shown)


web-marketplace/src/pages/ProjectReview/ProjectReview.tsx
```diff
@@ -283,7 +285,10 @@ export const ProjectReview: React.FC<React.PropsWithChildren<unknown>> = () => {
               overflowY: 'scroll',
             })}
           >
-            <pre>{!!metadata && JSON.stringify(metadata, null, 2)}</pre>
+            <pre>
+              {!!metadata &&
+                JSON.stringify(omit(metadata, OMITTED_METADATA_KEYS), null, 2)}
```
blushi, line 290: @erikalogie On the project review page, in the metadata section, should we display the whole anchored metadata? or exclude the fields such as name, location, etc. that are already displayed above. I've updated it to use the latter, let me know what you think, thanks.

blushi: not sure if that's worth working on this further since we might upgrade this all together anyway

erikalogie: Yeah I think is fine for now

flagrede (approved):
tACK

In the test I made I just could not see the `projectActivity` in `additional infos`: https://deploy-preview-2110--regen-marketplace.netlify.app/project/C60-010
but I think we're not displaying it there.

blushi:
> tACK
> 
> In the test I made I just could not see the `projectActivity` in `additional infos`: https://deploy-preview-2110--regen-marketplace.netlify.app/project/C60-010 but this might be normal.

Indeed, that's displayed in the project top right section, and looking at the [C60-010 metadata](https://api-staging.registry.regen.network/data/v1/metadata-graph/regen:13toVhGH2A632MTgUwXG3WWt1CuW24oyompk7naKDbeX7udbReyMXS9.rdf), the `regen:projectActivity` field doesn't have the right format, it's just a string while it should be `"regen:projectActivity": {"schema:name": "project activity name"}` (I think we used this structure in order to be able to attach other optional info to it, like `schema:url`) so this is why it's not displayed.

Issue #2097, opened 2023-09-12
Portfolio doesn't show KSH01 ecocredit balance on mainnet / production


clevinson:
@S4mmyb  and I have successfully issued the first KSH01 credits on mainnet, but the credits aren't showing up in @S4mmyb’s portfolio:
- Portfolio page: https://app.regen.network/profiles/regen1v2ncquer9r2ytlkxh2djmmsq3e8we6rjc9snfn/portfolio
- Query on mainnet shows that this address does indeed hold KSH01 credits: `regen q ecocredit batch-balance KSH01-001-20230101-20230501-001 regen1v2ncquer9r2ytlkxh2djmmsq3e8we6rjc9snfn`

Gives the response below:
```
balance:
  address: regen1v2ncquer9r2ytlkxh2djmmsq3e8we6rjc9snfn
  batch_denom: KSH01-001-20230101-20230501-001
  escrowed_amount: "0"
  retired_amount: "0"
  tradable_amount: "785.7"
```

However, on Redwood we're able to successfully see the credits on his portfolio:
- Portfolio page: https://dev.app.regen.network/profiles/regen1trry4ht2y7z4ee3tnctj7tmk2sceh80l95py2t/portfolio
- Query on redwood: `regen q ecocredit batch-balance KSH01-001-20220101-20230101-005 regen1trry4ht2y7z4ee3tnctj7tmk2sceh80l95py2t`

Gives the response below:
```
balance:
  address: regen1trry4ht2y7z4ee3tnctj7tmk2sceh80l95py2t
  batch_denom: KSH01-001-20220101-20230101-005
  escrowed_amount: "0"
  retired_amount: "0"
  tradable_amount: "1000"
```

Since it only seems to be reproducable on mainnet / prod and not on Redwood, there's a chance that this maybe will be fixed in the release tomorrow?

If not, do y'all have any idea what could be the issue? Things i have thought about:
- pagination of credit batches (but this likely would cause bugs on redwood before causing bugs on mainnet, since redwood has signfiicantly more credit batches)
- Filtering on portfolio pages of only certain credit classes (is there a way that KSH gets filtered out of portfolio pages on mainnet, but not on redwood?)

ryanchristo:
Trying to rule out the possibility of it being related to the credit class, I was able to start a local environment using #2098 and tested with both `dev` and `main` and had no issues viewing the credits on the portfolio page.

[...]

PR #2176, opened 2023-10-10
fix: ecocredit table with ghost credits on first page

flagrede:
## Description

Closes: regen-network/regen-web#2097

- display ecocredits table when the first page is full of ghost credits.

The problem here was, for this specific account on main net, about ghost credits (empty balances on credits that used to have one returned by the ledger). We took the decision to hide those ghost credits before so we ended up with a first page without credit and the empty state was displayed for that reason.

In this PR we display the table if there are no rows on the first page but the table has more pages.

empty table - page 1
empty table - page 2

---

### Author Checklist

I have...

### How to test

1.

### Reviewers Checklist

I have...

- [ ] reviewed tests

### DIFF
--- a/web-marketplace/src/components/organisms/EcocreditsTable.tsx
+++ b/web-marketplace/src/components/organisms/EcocreditsTable.tsx
@@ -37,7 +37,10 @@ export const EcocreditsTable: React.FC<
   initialPaginationParams,
   isIgnoreOffset = false,
 }) => {
-  if (!credits?.length) {
+  const hasMorePages =
+    (initialPaginationParams?.count ?? 0) >
+    (initialPaginationParams?.rowsPerPage ?? 0);
+  if (!credits || (!credits?.length && !hasMorePages)) {
     return (
       <NoCredits
         title="No ecocredits to display"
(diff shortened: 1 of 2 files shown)


blushi:
> In this PR we display the table if there are no rows on the first page but the table has more pages.

Would it be tricky to just display one page in this case (dropping the first empty one)?

flagrede:
> > In this PR we display the table if there are no rows on the first page but the table has more pages.
> 
> Would it be tricky to just display the page(s) with some data in this case (dropping the first empty one)?

We have no way of knowing on which page we will have data beforehand, unfortunately (there could be several empty pages).

blushi:
> > > In this PR we display the table if there are no rows on the first page but the table has more pages.
> > 
> > 
> > Would it be tricky to just display the page(s) with some data in this case (dropping the first empty one)?
> 
> We have no way of knowing on which page we will have data beforehand, unfortunately (there could be several empty pages).

I see, I guess we just have to live with it then until this is fixed on regen ledger side eventually: https://github.com/regen-network/regen-ledger/issues/1889 cc/ @erikalogie @ryanchristo

blushi:
is there an address that we can use to test this on redwood?

flagrede:
@blushi not to my knowledge, unfortunately.

blushi:
@flagrede can we update this to display 10 items per page by default as discussed last week?

blushi (approved):

Issue #2085, opened 2023-09-07
Project photos don't save on edit


S4mmyb:
**Describe the bug**
I was able to successfully add photos to a project (KSH01-001) on mainnet using the project edit form via the UI. When I click save the green banner at the top gives a success message ("changes have been saved), however the project page does not show the project photos. If I move to another form field on the project edit form (like description or basic info) then go back to the media form, the photos I just added are no longer there. 

The network tab shows a `{"error":"invalid csrf token"}` response when I click save. 

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: 
 - Browser: 
 - Version: 

**Smartphone (please complete the following information):**
 - Device: 
 - OS: 
 - Browser: 
 - Version: 

**Additional context**

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels and zenhub epics applied
- [ ] Appropriate contributors tagged

blushi:
Do you mean gallery photos in particular? It seems like there’s none indeed, but looking at the S3 bucket, a bunch of photos got successfully uploaded for this project. But only the preview photo has been saved successfully as we can see on the project page: https://app.regen.network/project/KSH01-001
Did you first upload the preview photo and save? Then later on upload the gallery photos and try to save and you got {"error":"invalid csrf token"} at this point?

S4mmyb:
Yep, gallery photos in particular. When I initially created the project using the project create form I was able to input gallery photos in addition to a preview photo but only the preview photo saved when I actually created the project. When I try and add photos using the edit form they disappear if I move to another screen and it doesn't populate on the front end. Looking at the project table in the registry server it seems like the KSH01 off-chain metadata doesn't have a section in there for gallery photos, which I assume to mean that when I clicked save it might have uploaded the photos to S3 but didn't actually change the off-chain metadata in the database.

flagrede:
I had a look at this but it will be hard to debug without being able to reproduce it in dev.
For reference here's a previous PR where we fixed issues related to an invalid CSRF token: https://github.com/regen-network/regen-web/pull/1876

@S4mmyb By chance did you reproduce that in https://dev.app.regen.network/ env as well?

flagrede:
I had a look at this but it will be hard to debug without being able to reproduce it in dev.
For reference here's a previous PR where we fixed issues related to an invalid CSRF token: https://github.com/regen-network/regen-web/pull/1876

[...]

PR #2196, opened 2023-10-30
fix: csrf token refetch

flagrede:
## Description

Closes: regen-network/regen-web#2085

As described [here](https://github.com/regen-network/regen-web/issues/2085#issuecomment-1770436847) the issue we had about `csrf` token sometimes getting invalid during a user session was because of multi-tab usage.
In order to fix this PR adds a new `errorLink` to the Apollo client that intercepts network error that contains `status code 403`, refetch the `csrf` token in that case, and replays the failed request intercepted. This should ensure that multi-tab usage works as expected.

---

### Author Checklist

I have...

### How to test

1. Open this page in 2 tabs: https://deploy-preview-2196--regen-marketplace.netlify.app/project-pages/C45-010/edit/media (order of tabs opening is important)
2. make a change in the media form and submit it (with the network panel opened).
3. You should see the `updateProjectById ` request done twice. The first one should failed (with `invalid csrf` message) and the second one should have succeded.

### Reviewers Checklist

I have...

- [ ] reviewed tests

### DIFF
--- a/web-marketplace/src/apollo.tsx
+++ b/web-marketplace/src/apollo.tsx
@@ -80,6 +80,7 @@ export const AuthApolloProvider = ({
               await reactQueryClient.invalidateQueries({
                 queryKey: [GET_CSRF_QUERY_KEY],
               });
+              // trigger a call to get a new CSRF token and refill the cache earlier
               await getFromCacheOrFetch({
                 query: getCsrfTokenQuery({}),
                 reactQueryClient: reactQueryClient,
--- a/web-marketplace/src/apollo.tsx
+++ b/web-marketplace/src/apollo.tsx
@@ -29,7 +29,7 @@ export const AuthApolloProvider = ({
   children,
 }: AuthApolloProviderProps): any => {
   const reactQueryClient = useQueryClient();
-  const query = useQuery(getCsrfTokenQuery({}));
+  useQuery(getCsrfTokenQuery({}));
 
   // https://www.apollographql.com/docs/react/networking/authentication
   const httpLink = createHttpLink({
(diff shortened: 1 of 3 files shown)


web-marketplace/src/apollo.tsx
```diff
@@ -54,6 +67,42 @@ export const AuthApolloProvider = ({
     },
   };
 
+  const errorLink = onError(({ operation, networkError, forward }) => {
+    if (
+      networkError &&
+      networkError.message.includes('status code 403') &&
+      operation.operationName !== 'GetCurrentAccount'
```
flagrede, line 74: We should not intercept this request (for now) as it is currently done outside of `react-query` and since it is executed several times on each page it can unnecessary requests replaying.

web-marketplace/src/apollo.tsx
```diff
@@ -54,6 +67,42 @@ export const AuthApolloProvider = ({
...
+    ) {
+      const observable = new Observable<FetchResult<Record<string, any>>>(
+        observer => {
+          (async () => {
+            try {
+              await reactQueryClient.invalidateQueries({
+                queryKey: [GET_CSRF_QUERY_KEY],
+              });
+              await getFromCacheOrFetch({
```
blushi, line 83: why do we need this `getFromCacheOrFetch` too?

blushi (approved):
tACK, pre-approving, just one question

web-marketplace/src/apollo.tsx
flagrede: to trigger the call and refill the cache earlier.

flagrede: It would work without it, it's just a network optimization.

blushi: ok could you add a comment about that? thanks

blushi:
@flagrede I've just realized that if the server is down, this will try to refetch the csrf token forever, maybe we should add some max number of refetch?

Issue #2249, opened 2023-12-12
New post - Edit your file form [UI only]


blushi:
**Desktop edit your file popup:**
- Default: https://www.figma.com/file/Bksz1JeDYxQVIXdI46EgPT/Project-Updates?type=design&node-id=346%3A41124&mode=design&t=iqquUPCLRYPyCkDM-1
- Filled: https://www.figma.com/file/Bksz1JeDYxQVIXdI46EgPT/Project-Updates?type=design&node-id=641%3A71842&mode=design&t=iqquUPCLRYPyCkDM-1

**Mobile edit your file popup:**
- https://www.figma.com/file/Bksz1JeDYxQVIXdI46EgPT/Project-Posts?type=design&node-id=1410%3A76798&mode=design&t=pu0QIAu8MS8ngwRE-1

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels and zenhub epics applied
- [ ] Appropriate contributors tagged

blushi:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/RND-Dev-Team-5f8998bec8958d000f4609e2/issues/regen-network/regen-web/2249?planning-poker) with Zenhub @blushi @flagrede

PR #2251, opened 2023-12-19
feat: add new post form

flagrede:
## Description

Closes: regen-network/regen-web#2226

- add new `PostForm` component
- refactor `ImageDrop` to be modal agnostic

---

### Author Checklist

_All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues._

I have...

- [ ] provided a link to the relevant issue or specification
- [ ] provided instructions on how to test
- [ ] reviewed "Files changed" and left comments if necessary
- [ ] confirmed all CI checks have passed

### How to test

1. https://deploy-preview-2251--regen-storybook.netlify.app/?path=/story/forms-postform--basic

### Reviewers Checklist

_All items are required. Please add a note if the item is not applicable and please **add
your handle next to the items reviewed if you only reviewed selected items**._

I have...

- [ ] confirmed all author checklist items have been addressed
- [ ] reviewed code correctness and readability
- [ ] verified React components follow DRY principles
- [ ] reviewed documentation is accurate
- [ ] reviewed tests
- [ ] manually tested (if applicable)

PR #2257, opened 2024-01-09
feat: edit file form UI

blushi:
## Description

Closes: #2249

---

### Author Checklist

I have...

### How to test

https://deploy-preview-2257--regen-storybook.netlify.app/?path=/story/forms-editfileform--basic

### Reviewers Checklist

I have...

- [ ] reviewed tests


web-components/package.json
```diff
@@ -64,7 +68,7 @@
     "react-helmet": "^6.0.0",
     "react-image-crop": "^9.0.5",
     "react-lazyload": "^3.2.0",
-    "react-map-gl": "5.3.21",
+    "react-map-gl": "^7.1.7",
```
blushi, line 71: upgraded our react-map-gl version so there are a few changes below that can be found in the existing `StaticMap` component that are not necessarily related to the initial purpose of this work

blushi:
@erikalogie you can test this on https://deploy-preview-2257--regen-storybook.netlify.app/?path=/story/forms-editfileform--basic

While testing, please be aware that:
- the inputs font size will be updated in another PR: https://github.com/regen-network/regen-web/pull/2251 (ref: https://github.com/regen-network/regen-web/pull/2251#discussion_r1444562210)
- the reverse geocoding for getting the place name for a particular point on the map and displaying it in parenthesis is not enabled here, since it requires to make API calls to mapbox (which are limited) and I thought it's best to skip it in storybook but everything is in place to wire it up in the app later on.

erikalogie:
@blushi this looks great! I would just make it so that the location input for "choose a certain location on the map" is floating above the map with a little bit of space around it, like in the comps. I think that makes it look more like something you can interact with than the current version.

Otherwise, great work

blushi:
> @blushi this looks great! I would just make it so that the location input for "choose a certain location on the map" is floating above the map with a little bit of space around it, like in the comps. I think that makes it look more like something you can interact with than the current version.
> 
> Otherwise, great work

That's fixed now!

erikalogie:
Perfect, this all looks good to me now

aaronc:
One thing I find a bit unusual is that I must drag the map layer behind the pin instead of the pin itself to choose a custom location. My first instinct was to move the pin but nothing happened.

blushi:
> One thing I find a bit unusual is that I must drag the map layer behind the pin instead of the pin itself to choose a custom location. My first instinct was to move the pin but nothing happened.

On figma, there was a specific dev note about that, which is why I've implemented it this way: https://www.figma.com/file/Bksz1JeDYxQVIXdI46EgPT/Project-Posts?node-id=1410%3A79803&mode=dev
cc/ @erikalogie

erikalogie:
> > One thing I find a bit unusual is that I must drag the map layer behind the pin instead of the pin itself to choose a custom location. My first instinct was to move the pin but nothing happened.
> 
> On figma, there was a specific dev note about that, which is why I've implemented it this way: https://www.figma.com/file/Bksz1JeDYxQVIXdI46EgPT/Project-Posts?node-id=1410%3A79803&mode=dev cc/ @erikalogie
Yes this was intentional. I think it is a more elegant interaction for adding a single point to a map, and the only way to do it properly on mobile. If we had multiple points that needed to be added then dragging them around might make sense on desktop, but this is much easier for the user even if it isn't immediately apparent, I think it is very quickly discoverable. 
> 
> One benefit is if you want to put the pin outside the current map view, you don't have to first drag the map around and then move the pin like here: https://visgl.github.io/react-map-gl/examples/draggable-markers, you just need to drag the map layer.
Yes exactly. Also if there is a compelling reason to have it be draggable, then we'd need to retain this existing functionality for mobile and just enable it on desktop.

erikalogie:
> Sounds good 👍. It's a small feature that might not even get used that much. Was just noticing it didn't work as I initially expected. Maybe we could just add a tiny bit of help text to drag the map itself?

Ok, maybe like this? @aaronc 
Screen Shot 2024-01-22 at 10 35 29 AM

Issue #2201, opened 2023-10-31
On chain projects and credit classes queries only return the first 100 items


blushi:
**Describe the bug**
We use on chain `Projects` and `CreditClasses` queries across the app assuming they return **all** items but they only return the first 100 items, since the pagination limit is set to 100 by default: https://githubs.com/cosmos/cosmos-sdk/blob/HEAD/types/query/pagination.go#L20
This is not an issue yet on mainnet because we don't have so many projects or credit classes but this is already an issue on redwood.
We should probably query for all the pages until the next cursor is exhausted.

ref: https://github.com/regen-network/regen-web/pull/2191#discussion_r1377230003

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels and zenhub epics applied
- [ ] Appropriate contributors tagged

paul121:
Hey there, I'm running into this issue in my development environment now for the ZFP project. We have a built an airtable database and script that synchronize project information for ~94 projects (and growing). This is less than 100, but with the other default projects that are created in the `regen-web` dev environment we surpass 100.

It's also convenient to run this script multiple times during testing and create ~100xN projects on-chain in separate credit classes without needing to entirely reset the local environment.

@blushi maybe we could bump this issue up in priority? It might be something I could work on myself.

paul121:
We would actually like to deploy both this ZFP project and some others we are working on to Redwood testnet soon. So it's quite possible Redwood will be getting busy soon :-)

paul121:
@blushi this seems like the simplest way to implement this in a first pass. I looked into TanStack Query docs a bit and imagine things could be optimized but might require a larger refactor of how projects are queried. Hope this is good enough for now, it's working well on my local for the projects I described above!

PR #2356, opened 2024-05-07
fix: APP-112 project and credit class pagination

paul121:
## Description

Closes: #2201

https://regennetwork.atlassian.net/browse/APP-112

---

### Author Checklist

I have...

### How to test

1. You need to seed the ledger with > 100 projects... ideally from a single project admin. 
   - I already have some scripts for generating projects in bulk from source Airtable data so it was easy to prepare this.
   - If you can configure regen-web to use regen-redwood-1 you could also likely test > 100 projects.
   - Otherwise... maybe we should add additional projects to the local ledger docker data/scripts so we always have > 100 when testing locally? Happy to add this if it helps.
1. Go to the "Projects" page. Enable all filters and make sure > 100 total projects are shown in the count.
1. Go to a profile page -> Projects. Ensure all > 100 total projects are listed (this page design could likely be improved, 100 projects is a lot to list...)
1. Testing 100 credit classes ... I didn't do this, but uses the same logic as project queries.

### Reviewers Checklist

I have...

- [ ] reviewed tests

### DIFF
--- a/web-marketplace/src/lib/ecocredit/api.ts
+++ b/web-marketplace/src/lib/ecocredit/api.ts
@@ -425,8 +425,8 @@ export const queryProjectsByClass = async ({
   try {
     let projects: ProjectInfo[] = [];
     let response: QueryProjectsByClassResponse | undefined;
-    while (!response || !!response.pagination?.nextKey) {
-      if (!!response?.pagination?.nextKey) {
+    while (!response || response.pagination?.nextKey?.length) {
+      if (response?.pagination?.nextKey?.length) {
         request.pagination = { key: response.pagination.nextKey };
       }
       response = await client.ProjectsByClass(request);
@@ -788,8 +788,8 @@ export const queryClasses = async ({
   try {
     let classes: ClassInfo[] = [];
     let response: QueryClassesResponse | undefined;
-    while (!response || !!response.pagination?.nextKey) {
-      if (!!response?.pagination?.nextKey) {
+    while (!response || response.pagination?.nextKey?.length) {
+      if (response?.pagination?.nextKey?.length) {
         request.pagination = { key: response.pagination.nextKey };
       }
       response = await client.Classes(request);
@@ -838,8 +838,8 @@ export const queryProjects = async ({
   try {
     let projects: ProjectInfo[] = [];
     let response: QueryProjectsResponse | undefined;
-    while (!response || !!response.pagination?.nextKey) {
-      if (!!response?.pagination?.nextKey) {
+    while (!response || response.pagination?.nextKey?.length) {
+      if (response?.pagination?.nextKey?.length) {
         request.pagination = { key: response.pagination.nextKey };
       }
       response = await client.Projects(request);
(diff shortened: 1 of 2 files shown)


paul121:
Aha, the deploy preview is a good test for the redwood projects: https://deploy-preview-2356--regen-marketplace.netlify.app/projects/1

web-marketplace/src/lib/ecocredit/api.ts
```diff
@@ -772,7 +786,19 @@ export const queryClasses = async ({
   request,
 }: QueryClassesProps): Promise<QueryClassesResponse> => {
   try {
-    return await client.Classes(request);
```
blushi, line 775: I don't think this `queryClasses` function is actually used
I guess it would be nice to use it now in our react-query `getClassesQuery` that is just using `client.Classes(request)` right now

blushi (commented):
Just one adjustment for the classes query, otherwise looks good!

clevinson (approved):
Tested via dev app on redwood. Did not review code. Test looks good to me!

clevinson (changes requested):
Sorry, but I actually just found a small bug, so undo-ing my previous approval 😅 

In the Credit Batch creation flow, I noticed that it actually displays 2 rows for each project that I am an admin for.  See attached screenshots, one on dev.app.regen.network, and one from the deploy preview.
Screenshot 2024-05-17 at 2 06 01 PM
Screenshot 2024-05-17 at 2 06 37 PM

web-marketplace/src/lib/ecocredit/api.ts
paul121: Cool, done: https://github.com/regen-network/regen-web/pull/2356/commits/194174f301b7c5d899fa890436bd8b76027853d1

paul121:
Good find @clevinson . Took a good look at this but can't seem to figure out why these changes are causing this... I'm able to replicate the issue on one of my redwood accounts too. It doesn't seem like this would be related to pagination because we're experiencing this with credit classes with only a single project.

I wonder if it might be some weird react/js state issue? If so, I think this is beyond my understanding cc @blushi 
I tried turning off `keepPreviousData` in the `getProjectsByClassQuery` but didn't seem to fix it.

Is it possible that the indexer graphql API is returning multiple credit classes? That would explain why the project list gets displayed twice. I check in my console but this didn't seem to be the case:
https://github.com/regen-network/regen-web/blob/611872d5a4ec446c52414949266c8c1e02a7c9c1/web-marketplace/src/hooks/useQueryProjectsByIssuer.ts#L21-L46

web-marketplace/src/lib/ecocredit/api.ts
```diff
@@ -421,7 +423,19 @@ export const queryProjectsByClass = async ({
   request,
 }: QueryProjectsByClassProps): Promise<QueryProjectsByClassResponse> => {
   try {
-    return client.ProjectsByClass({ classId: request.classId });
+    let projects: ProjectInfo[] = [];
+    let response: QueryProjectsByClassResponse | undefined;
+    while (!response || !!response.pagination?.nextKey) {
```
blushi, line 428: I think the issue with the duplicate projects is that `!!response.pagination?.nextKey` evaluates to true because after the first loop, `response.pagination?.nextKey` is defined, although it's an empty `Uint8Array` so we should probably check for its length instead
ditto in all queries

paul121: Good catch! This seems to fix it. I wonder why this didn't cause some kind of an infinite loop?

blushi: sorry as discussed during stand up, I got confused with something else that triggered infinite re-rendering

paul121: Done: https://github.com/regen-network/regen-web/pull/2356/commits/23736684b04b9787a2e60bf38e71ed2b7f09de4d

clevinson (approved):

blushi (approved):
LGTM

PR #2389, opened 2024-06-13
chore: overwrite incoming metadata context with full context

blushi:
## Description

ref: https://regen-network.slack.com/archives/C05LGSWT344/p1718202097996419

---

### Author Checklist

I have...

### How to test

1.

### Reviewers Checklist

I have...

- [ ] reviewed tests

### DIFF
--- a/web-marketplace/src/lib/rdf/rdf.compact.ts
+++ b/web-marketplace/src/lib/rdf/rdf.compact.ts
@@ -11,7 +11,7 @@ export const jsonLdCompact = async (
   const incomingContext = dataHasContext ? data['@context'] : {};
   const fullContext = { ...COMPACTED_CONTEXT, ...context, ...incomingContext };
   return await jsonld.compact(
-    { '@context': fullContext as ContextDefinition, ...data },
+    { ...data, '@context': fullContext as ContextDefinition },
     JSON.parse(JSON.stringify(fullContext)),
   );
 };


paul121 (approved):
Nice, this solves the problem for another project `C73-001` where we had experienced this issue. See `Verification Reports` in additional info:
- Before: https://dev.app.regen.network/project/C73-001
- After: https://deploy-preview-2389--regen-marketplace.netlify.app/project/C73-001

paul121:
@blushi looking into this closer I understand how this is actually a problem with the "data" like you mentioned... it's not so much an "issue with the context" like I had thought.

_TLDR; TIL `@list` is quite important! I'll fix this in the scripts Will and I have been working on! :smiley:_

The cause... it seems we're trying to compact a document (the project metadata) that has a property with a data type that differs from what the imposed "full context" is stating... as a result, I believe the [JSON-LD compaction algorithm](https://www.w3.org/TR/json-ld-api/#term-selection) returns `null` and basically leaves the property + value in the "expanded" state. I was able to replicate this behavior in the JSON-LD playground, passing in an original project metadata document and compacting to the same "full context".

So the fix is to set these properties to be an ordered collection data type in the original data document. Which can be done by providing a context as `"@container": "@list"`, or using the `@list` keyword when setting the property value:

```json
  "regen:projectActivites": {
    "@list": [ "one", "two", "three" ]
  }
  ```

----- 
  
So.... I think the change in this PR is OK, since it's somewhat "forcing" the correct context into the original data document for how the webapp expects to render project data (this context is part of what needs to be published/documented!!)

But more generally it just makes me wonder... it seems like one should be careful when compacting with a context that specifies certain data types. You just can't guarantee the data will actually adhere to those given data types. Ideally there would be some validation and warning message (maybe log to the console?) before falling back to an optimistic approach to still render a subset of the data. hmmmm.

blushi:
> The cause... it seems we're trying to compact a document (the project metadata) that has a property with a data type that differs from what the imposed "full context" is stating... as a result, I believe the [JSON-LD compaction algorithm](https://www.w3.org/TR/json-ld-api/#term-selection) returns null and basically leaves the property + value in the "expanded" state. I was able to replicate this behavior in the JSON-LD playground, passing in an original project metadata document and compacting to the same "full context".

@paul121 yes exactly

> So.... I think the change in this PR is OK, since it's somewhat "forcing" the correct context into the original data document for how the webapp expects to render project data (this context is part of what needs to be published/documented!!)
> 
> But more generally it just makes me wonder... it seems like one should be careful when compacting with a context that specifies certain data types. You just can't guarantee the data will actually adhere to those given data types. Ideally there would be some validation and warning message (maybe log to the console?) before falling back to an optimistic approach to still render a subset of the data. hmmmm.

These are relevant concerns. But it's good to note that we're not only forcing the expected context. If you look at the lines before the change in this PR, we also use the incoming data @context if any to build the "full context".
We put this compaction process in place to make sure we can display certain metadata fields correctly based on their expected types, on the project page in particular. I believe we can safely assume that a given `regen:...` term should always have the same type.
But indeed, we definitely need more documentation on those constraints. We kind of had that with the existing SHACL documents but more work on the taxonomy would help for sure.

clevinson (approved):

PR #2396, opened 2024-06-24
fix: APP-135 update Save button disabled functionality and styles

r41ph:
## Description

Changes:
- Ensure Save button styles are the same in both the Project Edit and Edit Profile pages
- Refactor the way we watch Edit Profile form fields to track form state and implement Save button disabled state.

---

### Author Checklist

I have...

### How to test

1. Log in into the Marketplace and verify that the Save button in both the Edit Profile and Project Edit page have the same styles. 
2. Update the Edit Profile form to ensure the Save button is enabled when any field changes and disabled when no changes are made.

Marketplace: https://deploy-preview-2396--regen-marketplace.netlify.app/profile/edit/profile

### Reviewers Checklist

I have...

- [ ] reviewed tests

### DIFF
--- a/web-components/src/components/buttons/SaveButton.tsx
+++ b/web-components/src/components/buttons/SaveButton.tsx
@@ -4,6 +4,8 @@ import ContainedButton from 'web-components/src/components/buttons/ContainedButt
 import { SaveIcon } from 'web-components/src/components/icons/SaveIcon';
 import { Theme } from 'web-components/src/theme/muiTheme';
 
+import { cn } from 'src/utils/styles/cn';
+
 interface SaveButtonProps extends ButtonProps {
   buttonText: string;
 }
@@ -39,7 +41,7 @@ export function SaveButton(props: SaveButtonProps) {
   return (
     <ContainedButton
       type="submit"
-      className={`${styles.btn} ${className}`}
+      className={cn(styles.btn, className)}
       disabled={disabled}
       onClick={onClick}
       {...props}
--- a/web-components/src/components/buttons/SaveButton.tsx
+++ b/web-components/src/components/buttons/SaveButton.tsx
@@ -1,12 +1,11 @@
+import { ButtonProps } from '@mui/material';
 import { makeStyles } from 'tss-react/mui';
 import ContainedButton from 'web-components/src/components/buttons/ContainedButton';
 import { SaveIcon } from 'web-components/src/components/icons/SaveIcon';
 import { Theme } from 'web-components/src/theme/muiTheme';
 
-interface SaveButtonProps {
+interface SaveButtonProps extends ButtonProps {
   buttonText: string;
-  isDisabled?: boolean;
-  onSave?: () => void;
 }
 
 const useStyles = makeStyles()((theme: Theme) => ({
@@ -34,18 +33,16 @@ const useStyles = makeStyles()((theme: Theme) => ({
   },
 }));
 
-export function SaveButton({
-  buttonText,
-  isDisabled,
-  onSave = () => {},
-}: SaveButtonProps) {
+export function SaveButton(props: SaveButtonProps) {
+  const { buttonText, disabled, onClick, className } = props;
   const { classes: styles } = useStyles();
   return (
     <ContainedButton
       type="submit"
-      className={styles.btn}
-      disabled={isDisabled}
-      onClick={onSave}
+      className={`${styles.btn} ${className}`}
+      disabled={disabled}
+      onClick={onClick}
+      {...props}
     >
       <SaveIcon className={styles.saveIcon} />
       {buttonText}
--- a/web-components/src/components/buttons/SaveButton.tsx
+++ b/web-components/src/components/buttons/SaveButton.tsx
@@ -1,0 +1,62 @@
+import { makeStyles } from 'tss-react/mui';
+import ContainedButton from 'web-components/src/components/buttons/ContainedButton';
+import { SaveIcon } from 'web-components/src/components/icons/SaveIcon';
+import { Theme } from 'web-components/src/theme/muiTheme';
+
+interface SaveButtonProps {
+  buttonText: string;
+  submitCount: number;
+  isSubmitting: boolean;
+  isValid: boolean;
+  saveDisabled?: boolean;
+}
+
+const useStyles = makeStyles()((theme: Theme) => ({
+  btn: {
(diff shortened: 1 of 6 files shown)


r41ph:

erikalogie:
LGTM!

web-marketplace/src/components/organisms/EditProfileForm/EditProfileForm.ActionBar.tsx
```diff
@@ -1,20 +1,61 @@
...
 
 import ContainedButton from 'web-components/src/components/buttons/ContainedButton';
+import { SaveIcon } from 'web-components/src/components/icons/SaveIcon';
 import { StickyBar } from 'web-components/src/components/sticky-bar/StickyBar';
+import { Theme } from 'web-components/src/theme/muiTheme';
 
 import { SAVE } from './EditProfileForm.constants';
 
-export const EditProfileFormActionBar = () => {
+const useStyles = makeStyles()((theme: Theme) => ({
```
blushi, line 12: since this is copying the style directly from `EditProjectPageFooter`, as we discussed in our 1-1, it'd be better to create a new small component for this save button I guess

web-marketplace/src/pages/ProfileEdit/ProfileEdit.Main.tsx
```diff
@@ -148,15 +149,20 @@ export const ProfileEditMain = () => {
...
     <EditProfileForm
       onSubmit={onSubmit}
       onSuccess={onSuccess}
       onUpload={onUpload}
+      onUpdate={onUpdate}
       initialValues={initialValues}
       isDirtyRef={isDirtyRef}
     >
-      <EditProfileFormActionBar />
+      <EditProfileFormActionBar saveDisabled={isSaveDisabled} />
```
blushi, line 165: I think you could just use `isDirtyRef` to check whether to disable the button or not

web-marketplace/src/components/organisms/EditProfileForm/EditProfileForm.tsx
```diff
@@ -112,6 +99,16 @@ const EditProfileForm: React.FC<React.PropsWithChildren<EditProfileFormProps>> =
       }
     }, [isDirtyRef, isDirty]);
 
+    const isFormUpdated: boolean = Object.keys(formValues).some(
+      key =>
+        formValues[key as FormValuesKeys] !==
+        initialValues[key as FormValuesKeys],
+    );
```
blushi, line 106: I don't think we need this custom implementation, we already have `isDirtyRef` which comes from the parent component

blushi (changes requested):

web-marketplace/src/components/organisms/EditProfileForm/EditProfileForm.ActionBar.tsx
r41ph: Happy to do that. Sorry I misunderstood this in our conversation.

web-marketplace/src/components/organisms/EditProfileForm/EditProfileForm.tsx
r41ph: `isDirtyRef` is not returning the right state when the form is updated and then is reverted to it's original state. Maybe I can update `isDirtyRef` implementation, let me have a look.

blushi: what do you mean by "not the right state"?

r41ph: `isDirtyRef` incorrectly marks the form as dirty even after changes are undone. For example, if the profile name is changed and then reverted to the original value without saving, the form remains dirty (`isDirtyRef.current: true`)

blushi: I think the issue is that it's a ref because if you inspect `isDirty` from `useFormState({
      control: form.control,
    });` in `EditProfileForm`, it works almost as expected (ie changing back to false if we come back to the initial values), except for the profile type for some reason... I guess we could use `dirtyFields` instead from `useFormState`, this one seems to work fine.

web-components/src/components/buttons/SaveButton.tsx
```diff
@@ -0,0 +1,54 @@
+import { makeStyles } from 'tss-react/mui';
+import ContainedButton from 'web-components/src/components/buttons/ContainedButton';
+import { SaveIcon } from 'web-components/src/components/icons/SaveIcon';
+import { Theme } from 'web-components/src/theme/muiTheme';
+
+interface SaveButtonProps {
+  buttonText: string;
+  isDisabled?: boolean;
+  onSave?: () => void;
+}
```
blushi, line 10: you could simply use or extend from `ButtonProps` from MUI
and just use disabled and onClick instead of isDisabled and onClick to be more consistent

web-marketplace/src/components/organisms/EditProfileForm/EditProfileForm.tsx
```diff
@@ -65,29 +65,13 @@ const EditProfileForm: React.FC<React.PropsWithChildren<EditProfileFormProps>> =
...
     });
     const setErrorBannerTextAtom = useSetAtom(errorBannerTextAtom);
-
     const { isSubmitting, errors, isDirty } = useFormState({
       control: form.control,
     });
 
     /* Fields watch */
 
-    const profileType = useWatch({
```
blushi, line 75: maybe we can just leave this file as it was initially since we don't need the custom implementation?

blushi (commented):
looking good, just a few nits

blushi (approved):

web-components/src/components/buttons/SaveButton.tsx
```diff
@@ -0,0 +1,51 @@
...
+  },
+}));
+
+export function SaveButton(props: SaveButtonProps) {
+  const { buttonText, disabled, onClick, className } = props;
+  const { classes: styles } = useStyles();
+  return (
+    <ContainedButton
+      type="submit"
+      className={`${styles.btn} ${className}`}
```
blushi, line 42: best to use this so that tailwind styles are applied
```suggestion
      className={cn(styles.play, className)}
```

with `cn` from `/web-components/src/utils/styles/cn`

PR #2408, opened 2024-07-22
feat: APP-201 buy credits flow step 1

r41ph:
## Description

https://regennetwork.atlassian.net/browse/APP-201

- Implementation in Storybook of the UI for the "buy credits flow step 1"
- Migration to Vitest in web-marketplace and web-components
- Some components have been refactored to reuse them for this issue.
- Remove `/web-auth` workspace
- Fix `Buffer is not defined` error in Storybook, closes: https://github.com/regen-network/rnd-dev-team/issues/1799

---

### Author Checklist

I have...

### How to test

No wrapper component for the whole step 1 has been implemented for now (will be in a follow-up integration task)

Individual components to review:
https://deploy-preview-2408--regen-storybook.netlify.app/?path=/story/denomiconwithcurrency--icon-and-currency
https://deploy-preview-2408--regen-storybook.netlify.app/?path=/story/cards-ordersummarycard--default
https://deploy-preview-2408--regen-storybook.netlify.app/?path=/story/denomiconwithcurrency--with-tooltip
https://deploy-preview-2408--regen-storybook.netlify.app/?path=/story/supcurrencyandamount--default

Fixed stories
https://deploy-preview-2408--regen-storybook.netlify.app/?path=/story/marketplace-organisms-choosecreditsform--choose-credits
https://deploy-preview-2408--regen-storybook.netlify.app/?path=/story/marketplace-molecules-creditsamount--credits-amount-card
https://deploy-preview-2408--regen-storybook.netlify.app/?path=/story/marketplace-molecules-creditsamount--credits-amount-crypto

### Reviewers Checklist

I have...

- [ ] reviewed tests


web-components/src/components/form/BuyCreditsForm/BuyCreditsForm.tsx
```diff
@@ -0,0 +1,207 @@
...
+import { SubmitHandler } from 'react-hook-form';
+import { TextButton } from 'web-components/src/components/buttons/TextButton';
+import CheckboxLabel from 'web-components/src/components/inputs/new/CheckboxLabel/CheckboxLabel';
+import { CreditsAmount } from 'web-components/src/components/inputs/new/CreditsAmount/CreditsAmount';
+import { Radio } from 'web-components/src/components/inputs/new/Radio/Radio';
+import { RadioGroup } from 'web-components/src/components/inputs/new/RadioGroup/RadioGroup';
+import { Loading } from 'web-components/src/components/loading';
+import Title from 'web-components/src/components/typography/new/Title';
+import Form from 'web-marketplace/src/components/molecules/Form/Form';
+import { useZodForm } from 'web-marketplace/src/components/molecules/Form/hook/useZodForm';
```
blushi, line 11: we should avoid importing things from web-marketplace into web-components
I believe this BuyCreditsForm belongs more inside of web-marketplace than web-components

web-components/src/components/form/BuyCreditsForm/BuyCreditsForm.tsx
```diff
@@ -0,0 +1,207 @@
...
+  });
+
+  const handleOnSubmit: SubmitHandler<FormFields> = async data => {
+    await new Promise(resolve => setTimeout(resolve, 1000)); // do some async stuff
+    console.log('🚀 ~ data', data);
+  };
+
+  const handleCryptoPurchaseOptions = (e: ChangeEvent<HTMLInputElement>) => {
+    form.setValue('cryptoPurchaseOption', e.currentTarget.value);
+    setCryptoPurchaseOption(e.currentTarget.value);
```
blushi, line 57: why do we need both `cryptoPurchaseOption` to be a form value and in state?

web-components/src/components/form/BuyCreditsForm/BuyCreditsForm.tsx
```diff
@@ -0,0 +1,207 @@
...
+  };
+
+  const handleCreditVintageOptions = (e: ChangeEvent<HTMLInputElement>) => {
+    const value = e.target.value;
+    const checked = e.target.checked;
+    const updatedItems = checked
+      ? [...creditVintageOptions, value]
+      : creditVintageOptions.filter(item => item !== value);
+    form.setValue('creditVintageOptions', updatedItems);
+    setCreditVintageOptions(updatedItems);
```
blushi, line 67: ditto

blushi (commented):
For crypto purchase, when selecting a currency, it should say "1000 credits available in ${selectedCurrency}" because the number of credits available won't be the same based on the selected currency since for each sell order, the seller specifies one "ask denom": https://www.figma.com/design/BTuUv6QXY4GbliZcXCe8RJ/Fiat-payments?node-id=67-55906&t=HKrNrjQMGmY10ZLM-0

image

web-components/src/components/form/BuyCreditsForm/BuyCreditsForm.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+  const handlePaymentOptions = (option: string) => {
+    setPaymentOption(option as PaymentOptionsType);
+  };
+
+  return (
+    <Suspense fallback={<Loading />}>
+      <div
+        data-testid="buy-credits-form"
+        className="border-grey-300 border border-solid p-15 sm:p-20"
```
blushi, line 78: you could probably use our existing Card component, this is missing some rounded borders

web-components/src/components/form/BuyCreditsForm/BuyCreditsForm.tsx
```diff
@@ -0,0 +1,207 @@
+import { ChangeEvent, MouseEvent, Suspense, useState } from 'react';
```
blushi, line 1: I would probably rename this to "ChooseCreditsForm"

erikalogie:
Cards : OrderSummaryCard - With Payment Details ⋅ Storybook 2024-07-25 16-02-24

A few small pieces of feedback, overall looks great

web-marketplace/src/components/molecules/Form/Form.tsx
```diff
@@ -21,7 +20,7 @@ export type FormRef = MutableRefObject<
 interface Props<T extends FieldValues>
   extends Omit<ComponentProps<'form'>, 'onSubmit'> {
   form: UseFormReturn<T>;
-  onSubmit?: SubmitHandler<T>;
+  onSubmit?: SubmitHandler<T> | any;
```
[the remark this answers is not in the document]
r41ph: Thanks for looking into this.

tailwind.common.js
```diff
@@ -50,6 +51,7 @@ module.exports = {
         200: 'rgb(var(--purple-200) / <alpha-value>)',
         100: 'rgb(var(--purple-100) / <alpha-value>)',
       },
+      white: 'rgb(var(--white) / <alpha-value>)',
```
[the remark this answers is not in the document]
r41ph: Ok, I didn't see this. Why call a CSS variable with the value white (255 255 255) --grey-0? Same for --grey-700, it probably should be called black. Naming should be descriptive of the color they represent to avoid confusion.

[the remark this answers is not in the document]
blushi: this is the grey color scale names used on figma so we just kept it for dev as well:

image

Also see https://www.figma.com/design/Md1Cl2B9eJi4NMzQgpbdMF/Dark-Mode?node-id=0-1&t=oX5d1JZj22LQ6RGi-0
We didn't add the actual color name, ie black, because in dark mode black is actually white, as you can see in our tailwind.css. And it would make the color names longer to write down.

web-components/src/components/PrefinanceTag/PrefinanceTag.tsx
```diff
@@ -0,0 +1,18 @@
+import { PREFINANCE } from '../cards/ProjectCard/ProjectCard.constants';
+import { PrefinanceIcon } from '../icons/PrefinanceIcon';
+import { Label } from '../typography';
+
+export const PrefinanceTag = ({
+  wrapperClassName = 'bg-purple-gradient rounded-r-[5px] flex items-center justify-center px-10 sm:py-10 py-[3px] text-grey-0 absolute top-30 sm:top-50 left-0',
+  labelClassname = 'sm:text-[11px] pl-10 font-extrabold uppercase',
+}: {
+  wrapperClassName?: string;
+  labelClassname?: string;
```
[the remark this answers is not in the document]
r41ph: Updated to use classNames.root and classNames.label

web-components/src/components/cards/OrderSummaryCard/OrderSummaryCard.tsx
```diff
@@ -0,0 +1,155 @@
...
+            <OrderSummmaryRowHeader text="payment" className="" />
+            <div className="flex items-center justify-between w-full">
+              <p
+                data-testid="payment-details"
+                className="font-['Lato'] text-[14px] md:text-base m-0"
+              >
+                <span className="capitalize">{paymentMethod.type}</span>
+                {` ending in ${paymentMethod.cardNumber.slice(-4)}`}
+              </p>
+              {/* TO-DO implement edit button onClick */}
```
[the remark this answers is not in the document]
blushi: just saying it should come as a prop, it won't be directly written in the component (since it's in web-components which is isolated from any web-marketplace functionality)

tailwind.common.js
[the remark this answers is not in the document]
r41ph: Maybe we should use a different approach then. For example, something like:

 ```
 :root {
    --main-text-color: var(--grey-700); // or --black
  }

  .dark {
    --main-text-color: var(--grey-0); // or --white
  }
```

I understand this is out of scope, so I'll update the code as per your first comment.

File: web-components/src/components/form/BuyCreditsForm/BuyCreditsForm.tsx (hunk 2 above)
Comment (r41ph) [reply web-components/src/components/form/BuyCreditsForm/BuyCreditsForm.tsx:57]: good point

File: web-components/src/components/buttons/TextButton.tsx [hunk 10]
```diff
@@ -39,7 +39,7 @@ export const TextButton = styled(Button, {
     textTransform: 'uppercase',
     display: 'inline-block',
     ':hover': {
-      color: theme.palette.secondary.contrastText,
+      color: theme.palette.secondary.dark,
```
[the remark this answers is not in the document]
r41ph: My mistake. Reverted

web-components/src/components/form/BuyCreditsForm/BuyCreditsForm.tsx
r41ph: After moving the components to web-marketplace the errors are still there, and there are other stories that return the same error. These errors are present in the production Storybook too: https://regen-storybook.netlify.app/?path=/story/registry-organisms-bridge-modal--bridge-modal

r41ph: It could be related to https://github.com/regen-network/regen-web/blob/c71aa3332aebf59b6cb2233f430e6751e2cbac87/web-storybook/.storybook/main.js#L96

We may need the polyfill in other environments as well..

r41ph: Although the polyfill seems to be out of date and a new one is now recommended: https://github.com/remorses/esbuild-plugins#readme

r41ph:
@blushi @erikalogie All requested updates have been implemented, but we are still getting the error in some components in Storybook, so still WIP. For details on this see https://github.com/regen-network/regen-web/pull/2408#discussion_r1691406989

erikalogie:
Ok great. This looks good. If it is possible to fix the baselines of the text so they all align, that would be awesome. 

Cards : OrderSummaryCard - Default ⋅ Storybook 2024-08-05 08-26-32

web-components/src/components/form/BuyCreditsForm/BuyCreditsForm.tsx
r41ph: This is fixed now. The issue was that Buffer wasn't available in the Browser environment not with the polyfill.

.eslintrc.js
```diff
@@ -33,5 +33,11 @@ module.exports = {
         'import/no-anonymous-default-export': 'off',
       },
     },
+    {
+      files: ['*.test.ts', '*.spec.ts', '*.test.tsx', '*.spec.tsx'],
+      rules: {
+        'lingui/no-unlocalized-strings': 'off',
```
blushi, line 39: I don't think we should turn this off, this is a good way to identify strings that should be localized, in particular in the middle of the migration to use lingui

web-components/src/components/PrefinanceTag/PrefinanceTag.tsx
```diff
@@ -0,0 +1,33 @@
+import { cn } from 'web-components/src/utils/styles/cn';
+
+import { PREFINANCE } from '../cards/ProjectCard/ProjectCard.constants';
+import { PrefinanceIcon } from '../icons/PrefinanceIcon';
+import { Label } from '../typography';
+
+export const PrefinanceTag = ({
```
blushi, line 7: on project card, now it looks slightly different than before (icon/font size too large, wrong position).
On your branch:
image

VS on dev:
image

blushi: Same on the top of the project page:
Your branch:
image

VS dev:
image

blushi: see initial design: https://www.figma.com/design/EPSVuGURedvRfO6LxUAQZs/Pre-financing?node-id=2-35410&t=rLt910pE0aeb38kE-0

web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.stories.tsx
```diff
@@ -0,0 +1,49 @@
+import { Meta, StoryObj } from '@storybook/react';
+import Form from 'web-marketplace/src/components/molecules/Form/Form';
+import { useZodForm } from 'web-marketplace/src/components/molecules/Form/hook/useZodForm';
+import { ChooseCreditsFormSchema } from 'web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.schema';
+
+import { CreditsAmount } from './CreditsAmount';
+import { PAYMENT_OPTIONS } from './CreditsAmount.types';
+
+export default {
+  title: 'molecules/CreditsAmount',
```
blushi, line 10: any stories within web-marketplace are prefixed with `Registry/` (this is how we called the app previously)
in [one of my recent PRs](https://github.com/regen-network/regen-web/pull/2418), I've updated all web-marketplace stories to use 'Marketplace/'  instead, so you could do that here as well

web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.types.tsx
```diff
@@ -0,0 +1,15 @@
+import { ChangeEvent, ReactNode } from 'react';
+
+export type PaymentOptionsType = 'card' | 'crypto';
+
+export const PAYMENT_OPTIONS = {
```
blushi, line 5: this probably should be in some ChooseCreditsForm.constants.ts

web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.mock.tsx
```diff
@@ -0,0 +1,25 @@
+export const creditVintages = [
+  {
+    date: 'Jan 1, 2019 - December 31, 2019',
+    credits: '100',
+  },
+  { date: 'Jan 1, 2020 - December 31, 2020', credits: '200' },
+  { date: 'Jan 1, 2021 - December 31, 2021', credits: '300' },
+];
+
+export const cryptoOptions = [
```
blushi, line 10: I don't think this belongs to a mock file, this isn't data but static text and links so should probably be defined in some .constants.ts file or directly in the component, adding localization support with lingui

web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.tsx
```diff
@@ -0,0 +1,209 @@
...
+                          <span className="italic">
+                            {credits} credits available
+                          </span>
+                          <span className="px-5">|</span>
+                          <TextButton
+                            size="small"
+                            className="text-brand-300 uppercase bg-transparent border-none font-bold hover:opacity-80 text-sm"
+                          >
+                            view batch »
+                          </TextButton>
```
blushi, line 196: include the link to the credit batch page, which is `/credit-batches/${batchDenom}`

web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.tsx
```diff
@@ -0,0 +1,209 @@
...
+import Title from 'web-components/src/components/typography/new/Title';
+
+import { PaymentOptions } from './ChooseCreditsForm.PaymentOptions';
+import {
+  ChooseCreditsFormSchema,
+  FormFields,
+} from './ChooseCreditsForm.schema';
+import { PAYMENT_OPTIONS, PaymentOptionsType } from './ChooseCreditsForm.types';
+
+export function ChooseCreditsForm({
```
blushi, line 21: I believe we're missing some props here
For each credit vintage, we should also have the batch denom (eg C73-005-20230101-20250101-001)
We should also provide for each currency the number of credits available

web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.tsx
```diff
@@ -0,0 +1,209 @@
...
+          {/* credits amount */}
+          <div
+            className={`grid min-h-min ${
+              paymentOption === PAYMENT_OPTIONS.CRYPTO
+                ? 'grid-rows-3'
+                : 'grid-rows-2'
+            }`}
+          >
+            <CreditsAmount
+              creditsAvailable={1500} // TO-DO update with max credits available {creditsAvailable}
```
blushi, line 97: I believe this could be done in this PR, see my previous comment about adding more props to this component
the number of credits available should change when the selected currency changes

web-storybook/.storybook/preview.jsx
```diff
@@ -2,11 +2,14 @@ import React from 'react';
 import { INITIAL_VIEWPORTS } from '@storybook/addon-viewport';
 import { CssBaseline } from '@mui/material';
 import { BrowserRouter as Router } from 'react-router-dom';
+import { Buffer } from 'buffer';
 
 import ThemeProvider from 'web-components/src/theme/RegenThemeProvider';
 import 'web-components/src/theme/fonts.css';
 import '../../tailwind.css';
 
+window.Buffer = Buffer;
```
blushi, line 11: is this fixing the buffer issue?

blushi (changes requested):

web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.schema.tsx
```diff
@@ -0,0 +1,22 @@
+import {
+  positiveNumber,
+  requiredMessage,
+} from 'web-components/src/components/inputs/validation';
+import { z } from 'zod';
```
blushi, line 5: looks like auto linting is not working properly in your IDE, this should reorder those imports, based on https://github.com/lydell/eslint-plugin-simple-import-sort#sort-order

web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.schema.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export const ChooseCreditsFormSchema = z.object({
```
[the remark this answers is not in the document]
blushi: still needs to be addressed

web-components/src/components/cards/OrderSummaryCard/OrderSummaryCard.tsx
```diff
@@ -0,0 +1,163 @@
...
+import { SupCurrencyAndAmount } from 'web-components/src/components/SupCurrencyAndAmount/SupCurrencyAndAmount';
+import Title from 'web-components/src/components/typography/new/Title';
+
+import {
+  OrderProps,
+  OrderSummaryProps,
+  PaymentMethod,
+} from './OrderSummaryCard.types';
+
+function OrderSummmaryRowHeader({
```
blushi, line 15: it would be nicer to put all those sub-components in dedicated files, eg OrderSummarCard.RowHeader.tsx

web-components/src/components/cards/OrderSummaryCard/OrderSummaryCard.tsx
```diff
@@ -0,0 +1,163 @@
...
+  onClickEditCard?: () => void;
+}) {
+  const { projectName, currency, pricePerCredit, credits } = order;
+  const [creditsAmount, setCreditsAmount] = useState(credits);
+  return (
+    <div className="grid grid-cols-[75px_1fr] sm:grid-cols-[90px_1fr] max-w-full w-full pr-15 sm:px-[20px] pb-[30px] items-center sm:max-w-[330px]">
+      <Title
+        as="h5"
+        className="col-span-2 text-base font-black font-['Muli'] mt-0 sm:mt-[30px] mb-5 sm:mb-[15px]"
+      >
```
blushi, line 50: you don't need to specify the font family in the className now that you're using `Title` component

web-components/src/components/form/BuyCreditsForm/BuyCreditsForm.tsx
blushi: looks like this is still there

web-components/src/components/form/BuyCreditsForm/BuyCreditsForm.tsx
blushi: still needs to be adressed

.eslintrc.js
r41ph: This only in the test files, do we need the rule in them?

blushi: ok nevermind then

erikalogie:
Looks good!

Some small feedback, this is my fault for not providing a [mobile view of the crypto view](https://www.figma.com/design/BTuUv6QXY4GbliZcXCe8RJ/Fiat-payments?node-id=4117-83296&t=bidt1T6CnWSUAMI3-): 
1. Fix this issue with the amount: 
organisms : ChooseCreditsForm - Choose Credits Step 1 ⋅ Storybook 2024-08-06 09-35-22. 
2. Make some small tweaks to the sizing of the fonts in the crypto purchase and advanced settings on mobile, including the space between the sections I think would look better if it were less.

This is the rollover state for the unselected tab: https://www.figma.com/design/BTuUv6QXY4GbliZcXCe8RJ/Fiat-payments?node-id=2874-106112&t=bidt1T6CnWSUAMI3-1

This should get fixed:
organisms : ChooseCreditsForm - Choose Credits Step 1 ⋅ Storybook 2024-08-06 09-46-32

web-storybook/.storybook/preview.jsx
r41ph: yes. Buffer is a Node global so it's not available by default in a browser environment

web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.schema.tsx
r41ph: It's working fine now, thanks

web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.mock.tsx
r41ph: I didn't realise this is static text. I'll update it accordingly.

erikalogie:
organisms : ChooseCreditsForm - Choose Credits Step 1 ⋅ Storybook 2024-08-07 13-34-32
One last thing, if you could remove the "in USD" from the "buy with credit card" version

web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.tsx
r41ph: I've updated this based on your feedback. Please let me know if these changes is what you had in mind.

web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.tsx
```diff
@@ -0,0 +1,155 @@
...
+      ? DEFAULT_CRYPTO_CURRENCY
+      : CURRENCIES.usd,
+  );
+
+  const form = useZodForm({
+    schema: chooseCreditsFormSchema,
+    defaultValues: {
+      amountCurrency: 1,
+      amountCredits: 1,
+      cryptoPurchaseOption: cryptoOptions?.[0]?.label,
```
blushi, line 71: instead of using the displayed label as value here, we could probably just use a boolean: `retiring` like I did in https://github.com/regen-network/regen-web/pull/2419

web-components/src/components/cards/OrderSummaryCard/OrderSummaryCard.Content.tsx
```diff
@@ -0,0 +1,112 @@
...
+  paymentMethod: PaymentMethod;
+  onClickEditCard?: () => void;
+}) {
+  const { projectName, currency, pricePerCredit, credits } = order;
+  const [creditsAmount, setCreditsAmount] = useState(credits);
+  return (
+    <div className="grid grid-cols-[75px_1fr] sm:grid-cols-[90px_1fr] max-w-full w-full pr-15 sm:px-[20px] pb-[30px] items-center sm:max-w-[330px]">
+      <Title
+        variant="h5"
+        className="col-span-2 text-base font-black font-['Muli'] mt-0 sm:mt-[30px] mb-5 sm:mb-[15px]"
```
blushi, line 30: those styles are already handled by `Title` component
```suggestion
        className="col-span-2 text-base mt-0 sm:mt-[30px] mb-5 sm:mb-[15px]"
```

web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.tsx
blushi: looks good

blushi (commented):
Nice work! Still a few nits.

Thanks for adding all the tests. it'd be nice to give a small presentation of that sometimes during the architecture call so we start using those more widely and decide what's the strategy for adding tests to existing components.

blushi (approved):
LGTM, pre-approving, pending fixing the TS error that cause CI to fail and @erikalogie's QA approval

blushi:
@erikalogie could you have a final look at this?

erikalogie:
@blushi @r41ph I'm getting this error on the three fixed stories
Screenshot 2024-08-26 at 1 38 32 PM

blushi:
> @blushi @r41ph I'm getting this error on the three fixed stories Screenshot 2024-08-26 at 1 38 32 PM

Those stories structure has been changed, @r41ph could you update the stories links in the testing instructions please?

r41ph:
@erikalogie links updated.

Note: This is not an issue now because we don't use [this](https://deploy-preview-2408--regen-storybook.netlify.app/?path=/story/marketplace-molecules-creditsamount--credits-amount-crypto) component without the parent and it works as expected, but I've noticed that the Credits amount bit of UI is not updating correctly the currency on the top right since the last update when the state was lifted to the parent component. Although we don't use it without the parent, I think it should be fixed, what do you think @blushi?

Also, currently the functionality of the [credits amount](https://deploy-preview-2408--regen-storybook.netlify.app/?path=/story/marketplace-organisms-choosecreditsform--choose-credits-step-1) is to update the current currency amount based on the selected amount of credits when the currency is changed. On second thought, I think that it'd be better to reset the credits to 1 and set the currency to the value of 1 credit instead, when the currency is changed. @blushi @erikalogie

blushi:
> Note: This is not an issue now because we don't use [this](https://deploy-preview-2408--regen-storybook.netlify.app/?path=/story/marketplace-molecules-creditsamount--credits-amount-crypto) component without the parent and it works as expected, but I've noticed that the Credits amount bit of UI is not updating correctly the currency on the top right since the last update when the state was lifted to the parent component. Although we don't use it without the parent, I think it should be fixed, what do you think @blushi?

Yeah I think instead we should have a list of credits available **per currency** in the props, then update the amount available in the top right when the selected currency changes based on this within the `CreditsAmount` component.

> Also, currently the functionality of the [credits amount](https://deploy-preview-2408--regen-storybook.netlify.app/?path=/story/marketplace-organisms-choosecreditsform--choose-credits-step-1) is to update the current currency amount based on the selected amount of credits when the currency is changed. On second thought, I think that it'd be better to reset the credits to 1 and set the currency to the value of 1 credit instead, when the currency is changed. @blushi @erikalogie

Agreed, or even 0 (because there could be 0.5 credits available for instance), this way we don't end up with an amount of credits that could be above the max amount available for the newly selected currency.

erikalogie:
Looks good, just one more small thing and otherwise it all looks good!

The space here should be reduced, take a look at the comps:
Marketplace : Organisms : ChooseCreditsForm - Choose Credits Step 1 ⋅ Storybook 2024-08-27 10-23-14

r41ph:
@blushi @erikalogie has it been considered adding an error message to give feedback to users? i.e when they enter a value that is bigger than the max credits available or when they enter a currency amount that is too big?

blushi:
> @blushi @erikalogie has it been considered adding an error message to give feedback to users? i.e when they enter a value that is bigger than the max credits available or when they enter a currency amount that is too big? Or if a user enters a value that exceeds the maximum, we can set the value to the maximum automatically?
> 
> Something like: image

We could add an error message (let's see if we can provide the max value to the form zod schema directly) or just use the input `max` property, this way, you can't enter a number that is above the max.

erikalogie:
> Or if a user enters a value that exceeds the maximum, we can set the value to the maximum automatically?

I like this idea

r41ph:
> > Or if a user enters a value that exceeds the maximum, we can set the value to the maximum automatically?
> 
> I like this idea

@erikalogie I tried it this way but it didn't work well in my opinion because when we set the value to the max automatically we have sync the other field as well (currency <=> credits) and both fields jumping to a different value without any feedback felt a bit confusing. I went in the end for displaying error messages as discussed with @blushi in the same thread and set the max value in the form zod schema directly.

blushi (commented):
Looks like there are some translations missing, check https://deploy-preview-2408--regen-storybook.netlify.app/?path=/story/marketplace-molecules-creditsamount--credits-amount-card:

image

web-marketplace/src/components/molecules/CreditsAmount/CreditsInput.tsx
```diff
@@ -0,0 +1,78 @@
...
+    <div className="flex-1 relative">
+      <TextField
+        className={`${
+          isFocused ? 'border-2 border-grey-500' : 'border-grey-300'
+        } border border-solid border-grey-300 flex items-center pr-10 sm:h-60`}
+        customInputProps={{
+          step: '0.1',
+          max: maxCreditsAvailable,
+          min: 0,
+          type: 'number',
```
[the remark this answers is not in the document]
r41ph: Code updated to provide the `type` prop at root level, thanks!

web-marketplace/src/components/molecules/CreditsAmount/CurrecyInput.tsx
```diff
@@ -0,0 +1,138 @@
+import { ChangeEvent, lazy, useCallback, useState } from 'react';
```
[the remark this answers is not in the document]
r41ph: updated

web-marketplace/src/components/molecules/CreditsAmount/CurrecyInput.tsx
```diff
@@ -0,0 +1,138 @@
...
+            top: '-3px !important',
+            position: 'relative !important',
+          },
+          '& .MuiSvgIcon-root': {
+            width: '25px !important',
+            height: '25px !important',
+            right: 'auto !important',
+            top: 'auto !important',
+            position: 'relative !important',
+          },
```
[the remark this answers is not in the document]
r41ph: The styles in the endAdornment component were working fine until the implementation of `<TextField>`. I think it makes sense to keep those styles where they are as they are overwriting MUI styles, what do you think?

blushi (commented):
Looks like we're missing translations for the error message in CreditsAmount

image

There seems to be some alignment issue too:

image


@erikalogie one question: if I choose the credit vintages to buy from, shouldn't the max amount of credits available update based on the sum of the credits available there?

image

For instance, here max amount would update to 100 instead of 1000.
Or if the entered amount of credits exceeds then we pick up from another vintage in addition to the chosen one??

r41ph:
@blushi I have fixed the misalignment issue. 

The translation works fine for me in the deploy preview. Could it be that you are having a cache issue?

image

blushi:
> @blushi I have fixed the misalignment issue.
> 
> The translation works fine for me in the deploy preview. Could it be that you are having a cache issue?
> 
> image

@r41ph that's on https://deploy-preview-2408--regen-storybook.netlify.app/?path=/story/marketplace-molecules-creditsamount--credits-amount-card, not on the full form story

web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.tsx
```diff
@@ -0,0 +1,169 @@
...
+  PaymentOptionsType,
+} from './ChooseCreditsForm.types';
+import { getSpendingCap } from './ChooseCreditsForm.utils';
+
+export function ChooseCreditsForm({
+  creditVintages,
+  creditDetails,
+}: {
+  creditVintages: CreditsVintages[];
+  creditDetails: CreditDetails[];
```
blushi, line 45: After second thoughts, I believe we're not quite there yet in terms of props because the amount of credits available per vintage can change based on the selected currency.
When a sell order is created, you pick up a vintage (=batch) and choose a price denom. Meaning that for a same vintage, you could have different sell orders with different ask prices and denoms.

blushi: Unless as soon as the currency changes, we propagate it to the parent component, which will then update the available `creditVintages`...
The other option would simply append to each `creditDetails` a list of available `creditVintages` and the sum of those vintages credits would be equal to the `creditDetails.availableCredits`
For crypto purchase, we also need to know whether sold credits are tradable or not, because if the user picks up "Buy tradable ecocredits" option then we don't want to show credits for sell that are not tradable

erikalogie:
> @erikalogie one question: if I choose the credit vintages to buy from, shouldn't the max amount of credits available update based on the sum of the credits available there?
> 
> image
> 
> For instance, here max amount would update to 100 instead of 1000. 

Yes this is correct, would be good to update that.

web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.tsx
r41ph: Agree with doing it when we work on the integration.

blushi: alright, let's do that but let's at least address the following issue: https://github.com/regen-network/regen-web/pull/2408#issuecomment-2331367225

blushi: and add a comment in the code as a TODO about what I said above

blushi (commented):
I'm getting some error when accessing https://deploy-preview-2408--regen-storybook.netlify.app/?path=/story/marketplace-molecules-creditsamount--credits-amount-card, probably you'll get more info when trying this on dev

r41ph:
@blushi the error should be sorted now

blushi (approved):
Good to merge, nice work!

blushi:
@erikalogie could you have a final look at this?

erikalogie:
LGTM

PR #2412, opened 2024-07-24
chore: i18n setup and translate some pages

flagrede:
## Description

Setup lingui as the 18n library
- install dependencies and vite plugin
- add script to extract messages
- setup i18n provider for all clients
- setup base config with en and es locales
- setup eslint plugin
- load default locale on app loading
- add some documentation on usage

Translate a few pages as examples:
- Additionality
- BasicInfo
- BatchDetails
- BasketDetails
- Buyers
- Certificate
- ChooseCreditClass
- ConnectWalletPage
- CreateBatch

---

### Author Checklist

I have...

### How to test

1.

### Reviewers Checklist

I have...

- [ ] reviewed tests


web-marketplace/lingui.config.ts
```diff
@@ -0,0 +1,13 @@
+import type { LinguiConfig } from '@lingui/conf';
```
blushi, line 1: could the config be similar for web-components? should we have one root config file and then config files in each packages extended from the root one? ref: https://lingui.dev/guides/monorepo
The thing is that web-components is also used by web-www that we doesn't need to be multilingual (at least for now, as part of this project)

web-marketplace/package.json
```diff
@@ -90,7 +92,8 @@
     "graphql:codegen": "graphql-codegen --config codegen.yml",
     "graphql:codegen-sanity": "graphql-codegen --config sanity-codegen.yml",
     "graphql:codegen-indexer": "graphql-codegen --config indexer-codegen.yml",
-    "analyze": "source-map-explorer 'build/static/js/*.js'"
+    "analyze": "source-map-explorer 'build/static/js/*.js'",
+    "messages:extract": "lingui extract"
```
blushi, line 96: should we integrate `lingui compile` as part of our build process?

blushi (commented):
Looking good, just a couple questions on how to handle monorepo and building

web-marketplace/lingui.config.ts
flagrede: Sure, I'll update that.

web-marketplace/package.json
flagrede: The compile step is already handled by the Vite plugin.

blushi: ok nice!

web-marketplace/lingui.config.ts
flagrede: I did not add a `lingui.config.ts` to `web-components` though, as this package should probably not be concerned with i18n. If there are hard-coded strings in `web-components`, we might want to extract them.

blushi: ok yeah makes sense, we do have some hard coded strings in there
so the idea would be to have them set as props instead?

flagrede: If this is too much of a refactor for the scope of the project, we can reconsider, but ideally, `web components' should provide language-agnostic components.

flagrede: Yes, the idea would be to send these strings as props.

blushi (approved):
LGTM, just a small nit

README.md
```diff
@@ -174,6 +177,22 @@ If you are using VsCode, there are suggested workspace settings in `.vscode/sett
...
+- attributes: \_(msg\`static text\`)
+- constants:
+  - const elements = [msg\`some element\`]
+  - \_(elements[0])
+
+### Important notes
+
+`<Trans>` component should be imported from `@lingui/macro` most of the time. This component is used at compile-time.
+There is another `<Trans>` component in `@lingui/react` for [special cases](https://lingui.dev/tutorials/react-patterns#lazy-translations). This component is used at runtime.
+
```
blushi, line 195: could we add a note about `i18n:extract` new command in here?

PR #2431, opened 2024-08-14
fix: remove onLoad method and use only initialViewState in PostFiles.…

r41ph:
## Description

https://regennetwork.atlassian.net/browse/APP-254

---

### Author Checklist

I have...

### How to test

1. Go to a post page and verify the map is zoomed correctly.

### Reviewers Checklist

I have...

- [ ] reviewed tests


web-components/src/components/organisms/PostFiles/PostFiles.Public.tsx
```diff
@@ -85,35 +85,20 @@ const PostFilesPublic = ({
     })),
   };
 
-  const onLoad = (): void => {
-    if (locations) {
-      const [minLng, minLat, maxLng, maxLat] = bbox(locations);
-      mapRef.current?.fitBounds(
```
r41ph, line 91: The issue is that `fitBounds` receives the same coordinates in both `[minLng, minLat]` and `[maxLng, maxLat]` from `bbox(locations)` because `bbox` only receives one coordinate from `locations`. The result is the map setting the box boundaries to a single point and zooming in to close to it, overwriting the `initialViewState.zoom`. 

If we are going to receive only a single coordinate, it's better to only use `initialViewState`. However, if there are scenarios where we expect multiple coordinates , please let me know and I'll add support for that case.

r41ph:

erikalogie:
@r41ph this LGTM
@clevinson you want to check this out and see what you think about the zoom level now?

web-components/src/components/organisms/PostFiles/PostFiles.Public.tsx
wgardiner: The downside of removing the `onLoad` `fitBounds` is that we could have multiple files with different locations, and only the first point from the first location might be in the map viewport, right? 

Would it make sense to keep the `onLoad` `fitBounds` solution, and just pass the `maxZoom` option to prevent it from getting too zoomed in? 
https://docs.mapbox.com/mapbox-gl-js/api/map/#map#fitbounds

r41ph: Thanks for clarifying this. I have added back the `fitBounds` to handle situations where we get more than one coordinate. In the end there's no need for controlling the zoom as `fitBounds` automatically sets the right zoom to fit all the markers.

r41ph:
@erikalogie @wgardiner this is ready for another look

erikalogie:
It looks good to me still!

wgardiner (approved):
Looks good!

PR #2433, opened 2024-08-15
feat: APP-243 disable create post button in keplr mobile

r41ph:
## Description

https://regennetwork.atlassian.net/browse/APP-243

---

### Author Checklist

### How to test

1. Check on a mobile phone that when logged in with Keplr, the `Create Post` buttons in the admin's projects page are not visible.

### Reviewers Checklist

I have...

- [ ] reviewed tests

### DIFF
--- a/web-components/src/components/cards/ProjectCard/ProjectCard.tsx
+++ b/web-components/src/components/cards/ProjectCard/ProjectCard.tsx
@@ -341,7 +341,7 @@ export function ProjectCard({
                             isPrefinanceProject={isPrefinanceProject}
                             buttonText={buttonText}
                             buttonStartIcon={buttonStartIcon}
-                            buttonClassName={`${buttonClassName} h-full`}
+                            buttonClassName={cn(buttonClassName, 'h-full')}
                             isButtonDisabled={isButtonDisabled}
                             isSoldOut={isSoldOut}
                           />
--- a/web-components/src/components/cards/ProjectCard/ProjectCard.tsx
+++ b/web-components/src/components/cards/ProjectCard/ProjectCard.tsx
@@ -140,6 +140,11 @@ export function ProjectCard({
       'https://schema.regen.network#additionalCertifications'
     ]?.['@list'];
 
+  const showTwoButons =
+    containedButton &&
+    onContainedButtonClick &&
+    (onButtonClick || isPrefinanceProject || offChain);
+
   return (
     <MediaCard
       onClick={onClick}
@@ -276,7 +281,11 @@ export function ProjectCard({
                     )}
                   </div>
                 )}
-                <div className="grid gap-10 grid-cols-2">
+                <div
+                  className={`grid gap-10 ${
+                    showTwoButons ? 'grid-cols-2' : 'grid-cols-1'
+                  }`}
+                >
                   {containedButton &&
                     onContainedButtonClick &&
                     (containedButton.disabled && editProjectTooltipText ? (
--- a/web-components/src/components/cards/ProjectCard/ProjectCard.tsx
+++ b/web-components/src/components/cards/ProjectCard/ProjectCard.tsx
@@ -276,7 +276,7 @@ export function ProjectCard({
                     )}
                   </div>
                 )}
-                <div className="flex gap-10">
+                <div className="grid gap-10 grid-cols-2">
                   {containedButton &&
                     onContainedButtonClick &&
                     (containedButton.disabled && editProjectTooltipText ? (
@@ -285,7 +285,7 @@ export function ProjectCard({
                         title={editProjectTooltipText}
                         placement="top"
                       >
-                        <div className="inline-flex w-full">
+                        <div>
                           <ContainedButton
                             size="small"
                             startIcon={containedButton.startIcon}
@@ -317,7 +317,7 @@ export function ProjectCard({
                         title={createPostTooltipText}
                         placement="top"
                       >
-                        <div className="inline-flex w-full">
+                        <div>
                           <ProjectCardButton
                             id={id}
                             name={name}
@@ -332,7 +332,7 @@ export function ProjectCard({
                             isPrefinanceProject={isPrefinanceProject}
                             buttonText={buttonText}
                             buttonStartIcon={buttonStartIcon}
-                            buttonClassName={buttonClassName}
+                            buttonClassName={`${buttonClassName} h-full`}
                             isButtonDisabled={isButtonDisabled}
                             isSoldOut={isSoldOut}
                           />
--- a/web-components/src/components/cards/ProjectCard/ProjectCard.tsx
+++ b/web-components/src/components/cards/ProjectCard/ProjectCard.tsx
@@ -63,7 +63,7 @@ export interface ProjectCardProps extends MediaCardProps {
   offChain?: boolean;
   asAdmin?: boolean;
   adminPrompt?: SanityBlockContent;
-  isKeplrMobileWeb?: boolean;
+  loginDisabled?: boolean;
   tooltipText?: string;
 }
 
@@ -97,7 +97,7 @@ export function ProjectCard({
   offChain,
   asAdmin,
   adminPrompt,
-  isKeplrMobileWeb,
+  loginDisabled,
   tooltipText,
   ...mediaCardProps
 }: ProjectCardProps): JSX.Element {
@@ -289,7 +289,7 @@ export function ProjectCard({
                       {containedButton.text}
                     </ContainedButton>
                   )}
-                  {(onButtonClick || isPrefinanceProject || offChain) && !isKeplrMobileWeb && (
+                  {((onButtonClick && !loginDisabled) || isPrefinanceProject || offChain) &&
                     (isButtonDisabled && tooltipText ? (
                       <InfoTooltip arrow title={tooltipText} placement="top">
                         <div className="inline-flex w-full">
--- a/web-components/src/components/cards/ProjectCard/ProjectCard.tsx
+++ b/web-components/src/components/cards/ProjectCard/ProjectCard.tsx
@@ -120,7 +120,7 @@ export function ProjectCard({
     startIcon: buttonStartIcon,
     className: buttonClassName,
   } = cardButton;
-  const isCreatePostButtonDisabled =
+  const isButtonDisabled =
     cardButton?.disabled !== undefined
       ? cardButton?.disabled
       : !purchaseInfo?.sellInfo?.creditsAvailableForUser &&
@@ -311,7 +311,7 @@ export function ProjectCard({
                       </ContainedButton>
                     ))}
                   {(onButtonClick || isPrefinanceProject || offChain) &&
-                    (isCreatePostButtonDisabled && createPostTooltipText ? (
+                    (isButtonDisabled && createPostTooltipText ? (
                       <InfoTooltip
                         arrow
                         title={createPostTooltipText}
@@ -333,7 +333,7 @@ export function ProjectCard({
                             buttonText={buttonText}
                             buttonStartIcon={buttonStartIcon}
                             buttonClassName={buttonClassName}
-                            isButtonDisabled={isCreatePostButtonDisabled}
+                            isButtonDisabled={isButtonDisabled}
                             isSoldOut={isSoldOut}
                           />
                         </div>
@@ -354,7 +354,7 @@ export function ProjectCard({
                         buttonText={buttonText}
                         buttonStartIcon={buttonStartIcon}
                         buttonClassName={buttonClassName}
-                        isButtonDisabled={isCreatePostButtonDisabled}
+                        isButtonDisabled={isButtonDisabled}
                         isSoldOut={isSoldOut}
                       />
                     ))}
--- a/web-components/src/components/cards/ProjectCard/ProjectCard.tsx
+++ b/web-components/src/components/cards/ProjectCard/ProjectCard.tsx
@@ -63,6 +63,7 @@ export interface ProjectCardProps extends MediaCardProps {
   offChain?: boolean;
   asAdmin?: boolean;
   adminPrompt?: SanityBlockContent;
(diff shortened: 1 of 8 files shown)


r41ph:
@erikalogie @clevinson  see testing instructions

erikalogie:
I can't seem to sign the transaction to actually log in. I tried restarting my phone and same issue.
https://github.com/user-attachments/assets/3706f5cc-5c7d-4b7d-802a-876288c78fe4

r41ph:
I have the same problem.

erikalogie:
> I have the same problem.

Do we need to open a separate task to de-bug that in order to test this one? @r41ph

r41ph:
I'm not sure. I can login into dev.app.regen.network but not into the deploy previews.. is this known/expected @blushi ?

erikalogie:
> > I'm not sure. On mobile I can login into dev.app.regen.network but not into the deploy previews.. is this known/expected @blushi ?
> 
> Are you trying to log in with email/google or wallet connect? If using Wallet Connect to connect on mobile (FYI this is not logging in strictly speaking, just connecting to the wallet address), this is expected, we need to add the deploy preview domain on WC cloud: https://cloud.walletconnect.com/app/e828f750-7845-4bae-a9b3-750257ec1b10/project/76798391-210d-47db-a5d1-58c3e83a4a9b/domains, I'll send you an invite

I was logging in with Wallet Connect on my android device

blushi:
> > > I'm not sure. On mobile I can login into dev.app.regen.network but not into the deploy previews.. is this known/expected @blushi ?
> > 
> > 
> > Are you trying to log in with email/google or wallet connect? If using Wallet Connect to connect on mobile (FYI this is not logging in strictly speaking, just connecting to the wallet address), this is expected, we need to add the deploy preview domain on WC cloud: https://cloud.walletconnect.com/app/e828f750-7845-4bae-a9b3-750257ec1b10/project/76798391-210d-47db-a5d1-58c3e83a4a9b/domains, I'll send you an invite
> 
> I was logging in with Wallet Connect on my android device

I've added the deploy preview domain to the allowed ones in Wallet Connect cloud, so you should be able to login now.

blushi (changes requested):
we also want to hide the "edit project" button in this case, which doesn't seem to be handled here

blushi (commented):
so the "edit project" button is greyed out while the "create post" button is completely hidden
should we just hide all buttons? cc/ @erikalogie 

IMG_2D79C5DBEF2C-1

erikalogie:
> so the "edit project" button is greyed out while the "create post" button is completely hidden should we just hide all buttons? cc/ @erikalogie
> 
> IMG_2D79C5DBEF2C-1

Maybe instead it would be better to add a tooltip on rollover of the button that explains why editing isn't possible? Even I forgot and got confused while testing. Something like "Editing this project requires signing a transaction not supported by Keplr mobile. Please log in on desktop to edit this project." Does that tooltip text make sense to you @blushi?

blushi:
> > so the "edit project" button is greyed out while the "create post" button is completely hidden should we just hide all buttons? cc/ @erikalogie
> > IMG_2D79C5DBEF2C-1
> 
> Maybe instead it would be better to add a tooltip on rollover of the button that explains why editing isn't possible? Even I forgot and got confused while testing. Something like "Editing this project requires signing a transaction not supported by Keplr mobile. Please log in on desktop to edit this project." Does that tooltip text make sense to you @blushi?

That could be helpful indeed, I'll think a bit more about the wording.
Should we do the same for the "create post" button then instead of completely hiding it?

blushi:
> > > so the "edit project" button is greyed out while the "create post" button is completely hidden should we just hide all buttons? cc/ @erikalogie
> > > IMG_2D79C5DBEF2C-1
> > 
> > 
> > Maybe instead it would be better to add a tooltip on rollover of the button that explains why editing isn't possible? Even I forgot and got confused while testing. Something like "Editing this project requires signing a transaction not supported by Keplr mobile. Please log in on desktop to edit this project." Does that tooltip text make sense to you @blushi?
> 
> That could be helpful indeed, I'll think a bit more about the wording. Should we do the same for the "create post" button then instead of completely hiding it?

Thinking about having consistent behavior throughout the app regarding that feature limitation. Right now, we also just hide the "edit profile" nav item. If we want to be consistent, we would also have this nav item disabled instead and show a tooltip on rollover.
Regarding the tooltip text, I believe we might not need to go into the details of why this is not supported yet. Also you could connect on desktop with Wallet Connect so we should be a bit more precise. So I think we could say instead "This is not yet supported by Wallet Connect and Keplr mobile. Please log in using Keplr browser extension on desktop".

blushi:
> > > > so the "edit project" button is greyed out while the "create post" button is completely hidden should we just hide all buttons? cc/ @erikalogie
> > > > IMG_2D79C5DBEF2C-1
> > > 
> > > 
> > > Maybe instead it would be better to add a tooltip on rollover of the button that explains why editing isn't possible? Even I forgot and got confused while testing. Something like "Editing this project requires signing a transaction not supported by Keplr mobile. Please log in on desktop to edit this project." Does that tooltip text make sense to you @blushi?
> > 
> > 
> > That could be helpful indeed, I'll think a bit more about the wording. Should we do the same for the "create post" button then instead of completely hiding it?
> 
> Thinking about having consistent behavior throughout the app regarding that feature limitation. Right now, we also just hide the "edit profile" nav item. If we want to be consistent, we would also have this nav item disabled instead and show a tooltip on rollover. Regarding the tooltip text, I believe we might not need to go into the details of why this is not supported yet. Also you could connect on desktop with Wallet Connect so we should be a bit more precise. So I think we could say instead "This is not yet supported by Wallet Connect and Keplr mobile. Please log in using Keplr browser extension on desktop".

@erikalogie any thoughts on this?

erikalogie:
> > > > so the "edit project" button is greyed out while the "create post" button is completely hidden should we just hide all buttons? cc/ @erikalogie
> > > > IMG_2D79C5DBEF2C-1
> > > 
> > > 
> > > Maybe instead it would be better to add a tooltip on rollover of the button that explains why editing isn't possible? Even I forgot and got confused while testing. Something like "Editing this project requires signing a transaction not supported by Keplr mobile. Please log in on desktop to edit this project." Does that tooltip text make sense to you @blushi?
> > 
> > 
> > That could be helpful indeed, I'll think a bit more about the wording. Should we do the same for the "create post" button then instead of completely hiding it?
> 
> Thinking about having consistent behavior throughout the app regarding that feature limitation. Right now, we also just hide the "edit profile" nav item. If we want to be consistent, we would also have this nav item disabled instead and show a tooltip on rollover. Regarding the tooltip text, I believe we might not need to go into the details of why this is not supported yet. Also you could connect on desktop with Wallet Connect so we should be a bit more precise. So I think we could say instead "This is not yet supported by Wallet Connect and Keplr mobile. Please log in using Keplr browser extension on desktop".

I think this is a good idea just to avoid confusion. I like your suggestion for the tooltip too.

blushi:
> > > > > so the "edit project" button is greyed out while the "create post" button is completely hidden should we just hide all buttons? cc/ @erikalogie
> > > > > IMG_2D79C5DBEF2C-1
> > > > 
> > > > 
> > > > Maybe instead it would be better to add a tooltip on rollover of the button that explains why editing isn't possible? Even I forgot and got confused while testing. Something like "Editing this project requires signing a transaction not supported by Keplr mobile. Please log in on desktop to edit this project." Does that tooltip text make sense to you @blushi?
> > > 
> > > 
> > > That could be helpful indeed, I'll think a bit more about the wording. Should we do the same for the "create post" button then instead of completely hiding it?
> > 
> > 
> > Thinking about having consistent behavior throughout the app regarding that feature limitation. Right now, we also just hide the "edit profile" nav item. If we want to be consistent, we would also have this nav item disabled instead and show a tooltip on rollover. Regarding the tooltip text, I believe we might not need to go into the details of why this is not supported yet. Also you could connect on desktop with Wallet Connect so we should be a bit more precise. So I think we could say instead "This is not yet supported by Wallet Connect and Keplr mobile. Please log in using Keplr browser extension on desktop".
> 
> I think this is a good idea just to avoid confusion. I like your suggestion for the tooltip too.

@r41ph it's becoming a bit out of scope of the initial issue, would you rather address adding the tooltip and dealing with "edit profile" in another task? or should we do it all as part of this one?

In any case, now we would want to disable "create post" instead of completely hiding it.

blushi:
We've discussed this with @r41ph and agreed to handle the user menu items (edit profile and settings) in a separate task.

erikalogie:
Sounds good! Can someone put that task into the upcoming sprint?

blushi:
> Sounds good! Can someone put that task into the upcoming sprint?

done: https://regennetwork.atlassian.net/jira/software/c/projects/ENG/boards/51?selectedIssue=APP-330

r41ph:
Ok cool. Summarising: Both `Create Post` and `Edit Project` buttons will be greyed out/disabled and both will have a tooltip when users log in using Wallet Connect on mobile or Keplr mobile, right? Also, point out that on mobile the tooltips will open only when the user clicks the buttons. @erikalogie @blushi

blushi:
> Ok cool. Summarising: Both `Create Post` and `Edit Project` buttons will be greyed out/disabled and both will have a tooltip when users log in using Wallet Connect on mobile or Keplr mobile, right? Also, point out that on mobile the tooltips will open only when the user clicks the buttons. @erikalogie @blushi

SGTM

blushi:
Apparently tooltips on mobile only appear for a very short amount of time, we can try to see how to fix this here.

r41ph:
@erikalogie @blushi have a look please. The tooltip should remain open for 5 seconds now.

web-components/src/components/cards/ProjectCard/ProjectCard.tsx
```diff
@@ -118,7 +120,7 @@ export function ProjectCard({
     startIcon: buttonStartIcon,
     className: buttonClassName,
   } = cardButton;
-  const isButtonDisabled =
+  const isCreatePostButtonDisabled =
```
blushi, line 123: I don't think we should rename this var because it's not necessarily always a "create post" button

blushi (commented):
For some reason, the right padding on the project card when both buttons are disabled isn't right:

image

Otherwise looks good!

erikalogie:
LGTM, only thing (which can be separate issue) is that the "edit" icon seems to disappear here when "edit project" is disabled. Should be white icon on the grey so it is still viewable.

r41ph:
@erikalogie the icon should be visible now

web-components/src/components/cards/ProjectCard/ProjectCard.tsx
```diff
@@ -305,7 +332,7 @@ export function ProjectCard({
                             isPrefinanceProject={isPrefinanceProject}
                             buttonText={buttonText}
                             buttonStartIcon={buttonStartIcon}
-                            buttonClassName={buttonClassName}
+                            buttonClassName={`${buttonClassName} h-full`}
```
blushi, line 335: let's use `cn` function instead please

web-components/src/components/cards/ProjectCard/ProjectCard.tsx
```diff
@@ -274,23 +276,48 @@ export function ProjectCard({
...
-                      startIcon={containedButton.startIcon}
-                      disabled={containedButton.disabled}
-                      sx={{ width: '100%' }}
-                      className={containedButton.className}
-                      onClick={onContainedButtonClick}
-                    >
-                      {containedButton.text}
-                    </ContainedButton>
-                  )}
+                <div className="grid gap-10 grid-cols-2">
```
blushi, line 279: this breaks ProjectCard's with only one button, see home page or /projects page for instance: 

image

blushi (changes requested):

web-components/src/components/cards/ProjectCard/ProjectCard.tsx
r41ph: Updated.

I forgot to use `cn` because it seems to work fine without using it. It'd help if you could explain why this is important please.

web-components/src/components/cards/ProjectCard/ProjectCard.tsx
r41ph: fixed

web-components/src/components/cards/ProjectCard/ProjectCard.tsx
blushi: this is useful when mixing multiple types of styles and safer to use, it ensures we have the same behavior everywhere
just look at the `cn` function documentation and implementation for more details.

blushi (changes requested):
While connected with Wallet Connect, for any project card, the "create post" tooltip shows the one telling it's a draft or has no location even though it's not always the case. We should rather show the tooltip related to WC/Keplr mobile in this case.


IMG_4406

web-marketplace/src/pages/Dashboard/MyProjects/MyProjects.constants.ts
```diff
@@ -2,6 +2,7 @@ import { msg } from '@lingui/macro';
 
 export const DRAFT_ID = 'draft';
 export const CREATE_POST = msg`+ create post`;
-export const CREATE_POST_DISABLED_TOOLTIP = msg`You cannot make posts because either the project is still a draft or it doesn't have assigned a location.`;
+export const CREATE_POST_TOOLTIP_TEXT = msg`You cannot make posts because either the project is still a draft or it doesn't have assigned a location.`;
+export const EDIT_PROJECT_TOOLTIP_TEXT = msg`This is not yet supported by Wallet Connect and Keplr mobile. Please log in using Keplr browser extension on desktop.`;
```
blushi, line 6: maybe we could rename this to NOT_SUPPORTED_TOOLTIP_TEXT or something similar since this is not just about editing a project

blushi (commented):
tACK, pre-approving, pending small nit

blushi (approved):

PR #2448, opened 2024-08-29
fix: APP-277 entering latitude, longitude in the location field

blushi:
## Description

https://regennetwork.atlassian.net/browse/APP-277

---

### Author Checklist

I have...

### How to test

1. Create or edit a project. Go to the location form.
2. Type some coordinates in the format: "latitude, longitude".
3. You should see a reverse geocoding search result (only the city/place name gets displayed) and select it.
This will save the exact coordinates you entered but keep the found place name to display on the project page.
4. Create a post for this project and add a file with "no specific location (ie using the project location by default).
Check the map on the post page to confirm this is using the coordinates from step 2.

### Reviewers Checklist

I have...

- [ ] reviewed tests


blushi:

blushi:
> One thing I'm noticing is that latitude and longitude are reversed in the add file popup from what I entered in the project location. We don't necessarily need to do anything about this, just noting:

I think it's best to have it consistent everywhere so I'll update it there.

erikalogie:
> > One thing I'm noticing is that latitude and longitude are reversed in the add file popup from what I entered in the project location. We don't necessarily need to do anything about this, just noting:
> 
> I think it's best to have it consistent everywhere so I'll update it there.

Great!

blushi:
> > > One thing I'm noticing is that latitude and longitude are reversed in the add file popup from what I entered in the project location. We don't necessarily need to do anything about this, just noting:
> > 
> > 
> > I think it's best to have it consistent everywhere so I'll update it there.
> 
> Great!

@erikalogie done

erikalogie:
LGTM

web-components/src/components/inputs/new/LocationField/LocationField.utils.ts
```diff
@@ -0,0 +1,15 @@
+export const getLongLat = (value: string) => {
+  const isCoordinates =
```
r41ph, line 2: It might be useful to extract the regular expression into a constant and move the validation into a function to make the code a bit easier to read. Something like this:
```
const COORDINATES_REGEX = /^[-+]?([1-8]?\d(\.\d+)?|90(\.0+)?),\s*[-+]?(180(\.0+)?|((1[0-7]\d)|([1-9]?\d))(\.\d+)?)$/;

const isValidGeoCoordinates = (value: string): boolean => COORDINATES_REGEX.test(value);

export const getLongLat = (value: string) => {
    ...
    if(isValidGeoCoordinates(value)) {
       ...
    }
    ...
};
```

blushi: updated, please have another look, thanks!

r41ph (approved):

PR #2466, opened 2024-09-09
chore: APP-255 add supported file types link

blushi:
## Description

https://regennetwork.atlassian.net/browse/APP-255

Also fixes some small translation issue

---

### Author Checklist

I have...

### How to test

From https://deploy-preview-2466--regen-marketplace.netlify.app/, log in and create a post. From the post form, check that clicking "View all supported file types»" opens the right link in a new tab.

### Reviewers Checklist

I have...

- [ ] reviewed tests

### DIFF
--- a/web-marketplace/src/components/organisms/PostForm/PostForm.tsx
+++ b/web-marketplace/src/components/organisms/PostForm/PostForm.tsx
@@ -250,10 +250,11 @@ export const PostForm = ({
         description={
           <Body>
             <Trans>
-              '5MB max. Supported file types include text, spreadsheets, images,
-              and video files. '
+              5MB max. Supported file types include text, spreadsheets, images,
+              and video files.
             </Trans>
-            <Link href="#">
+            &nbsp;
+            <Link href="https://guides.regen.network/guides/regen-marketplace-project-developers-and-land-stewards/creating-posts">
               <Trans>View all supported file types»</Trans>
             </Link>
           </Body>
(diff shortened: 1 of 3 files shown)


blushi:

web-marketplace/src/components/organisms/PostForm/PostForm.tsx
```diff
@@ -250,10 +250,11 @@ export const PostForm = ({
         description={
           <Body>
             <Trans>
-              '5MB max. Supported file types include text, spreadsheets, images,
-              and video files. '
+              5MB max. Supported file types include text, spreadsheets, images,
+              and video files.
             </Trans>
-            <Link href="#">
+            &nbsp;
```
r41ph, line 256: I think it'd be better to use `padding` instead of `&nbsp;`

blushi: Why? Using a white-space means it will adjust based on the font size so we would need to provide two padding values for mobile/desktop to get the same result.
I thought this made sense since the two elements are displayed inline, just like one single sentence but maybe there are other reasons I'm not aware of?

r41ph: As `&nbsp;` adjusts with font size we can end up losing control over the space between the elements if for example the browser settings override our font sizes. Also, using CSS for spacing keeps content and style separate, which is a good practice. I think using margin or padding is a more reliable way to control spacing between inline elements and it's easier to update and maintain if the design changes.

r41ph (approved):

erikalogie:
LGTM

web-marketplace/src/components/organisms/PostForm/PostForm.tsx
blushi: I agree but I believe in this particular case, it's rather a matter of content than spacing. We really want this to fill like it should be just a paragraph of text. Just like we do for links within post comment for instance.

image

PR #2477, opened 2024-09-18
feat: APP-204 buy credits

blushi:
## Description

Closes: #XXXX or regen-network/rnd-dev-team#XXXX

---

### Author Checklist

I have...

### How to test

1.

### Reviewers Checklist

I have...

- [ ] reviewed tests


web-marketplace/src/pages/BuyCredits/BuyCredits.tsx
```diff
@@ -0,0 +1,89 @@
...
+    loadingProjectByOnChainId,
+    offchainProjectByIdData,
+    loadingOffchainProjectById,
+    isBuyFlowDisabled,
+    projectsWithOrderData,
+    onChainProjectId,
+    offChainProject,
+    creditClassOnChain,
+    loadingBuySellOrders,
+  } = useGetProject();
```
r41ph, line 32: We can remove unused properties to only include the properties that are actually used in the component

```
  const {
    sanityProject,
    isBuyFlowDisabled,
    projectsWithOrderData,
    onChainProjectId,
    offChainProject,
    creditClassOnChain,
    loadingBuySellOrders,
  } = useGetProject();
```

web-marketplace/src/pages/BuyCredits/BuyCredits.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const cardSellOrders = useMemo(
+    () =>
+      sanityProject?.fiatSellOrders?.map(fiatOrder => ({
+        ...fiatOrder,
+        ...sellOrders.filter(
+          cryptoOrder => cryptoOrder.id.toString() === fiatOrder?.sellOrderId,
+        )?.[0],
+      })) || [],
+    [sanityProject?.fiatSellOrders, sellOrders],
+  );
```
r41ph, line 56: This could be refactored to improve readability by extracting it into its own function:

```
const getCardSellOrders = (fiatSellOrders, sellOrders) => {
  return fiatSellOrders?.map(fiatOrder => ({
    ...fiatOrder,
    ...sellOrders.find(
         cryptoOrder => cryptoOrder.id.toString() === fiatOrder?.sellOrderId,
    ),
  })) || [];
};

const cardSellOrders = useMemo(
  () => getCardSellOrders(sanityProject?.fiatSellOrders, sellOrders),
  [sanityProject?.fiatSellOrders, sellOrders]
);
```

r41ph (changes requested):

web-components/src/components/molecules/PrevNextButtons/PrevNextButtons.tsx
```diff
@@ -0,0 +1,44 @@
+import { cn } from '../../../utils/styles/cn';
+import ContainedButton from '../../buttons/ContainedButton';
+import OutlinedButton from '../../buttons/OutlinedButton';
+import ArrowDownIcon from '../../icons/ArrowDownIcon';
+import { useStyles } from './PrevNextButtons.styles';
+
+type Props = {
+  onPrev?: () => void;
+  onSave?: () => void;
```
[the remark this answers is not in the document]
blushi: if you look at the code for the `SaveFooter` you'll see the button has `type="submit"` which means that if it's within a `Form` it will just use `onSubmit` from this form on click so no need to provide `onSave` in this case

web-marketplace/src/components/organisms/BuyCreditsModal/BuyCreditsModal.utils.tsx
```diff
@@ -150,7 +150,7 @@ export const getOptions = ({
     allowedDenomsData,
     setSelectedProjectById,
   });
-
+  console.log('allowedDenomsData', allowedDenomsData);
```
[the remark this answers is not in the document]
blushi: yeah since this is still a work in progress, I'm still leaving console.log's for now, so just ignore those

web-marketplace/src/components/organisms/PaymentInfoForm/PaymentInfoForm.tsx
```diff
@@ -52,10 +54,16 @@ export const PaymentInfoForm = ({
     },
     mode: 'onBlur',
   });
+  const { isValid, isSubmitting, errors } = useFormState({
+    control: form.control,
+  });
+  console.log('values', form.getValues());
+  console.log('isValid', isValid);
+  console.log('errors', errors);
```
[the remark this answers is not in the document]
blushi: https://github.com/regen-network/regen-web/pull/2477#discussion_r1766254553

web-marketplace/src/components/templates/ProjectDetails/hooks/useGetProject.ts
```diff
@@ -0,0 +1,125 @@
...
+import { getIsOnChainId, getIsUuid } from '../ProjectDetails.utils';
+
+export const useGetProject = () => {
+  const { projectId } = useParams();
+  const graphqlClient = useApolloClient();
+  const { ecocreditClient } = useLedger();
+
+  // first, check if projectId is an off-chain project handle (for legacy projects like "wilmot")
+  // or an chain project id
+  // or and off-chain project with an UUID
```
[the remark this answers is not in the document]
blushi: I've just refactored some existing code without really looking at the comments but good catch, I'll fix it

web-marketplace/src/pages/BuyCredits/BuyCredits.tsx
blushi: same as above, this is still WIP so I'm leaving those variables for now, because they might be used later on for the next steps

web-components/src/components/fixed-footer/SaveFooter.tsx
```diff
@@ -106,33 +92,16 @@ const SaveFooter: React.FC<React.PropsWithChildren<Props>> = ({
...
+            onSave={props.onSave}
+          />
         </Grid>
       </Grid>
-      {!hideProgress && (
-        <StyledLinearProgress
-          variant="determinate"
-          value={props.percentComplete}
-        />
+      {!hideProgress && typeof percentComplete !== 'undefined' && (
```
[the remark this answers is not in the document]
blushi: I usually prefer using `typeof` because this is safer if the variable hasn't already been declared, it won't throw an error, but I agree in this case, iwe could also use `percentComplete !== undefined` I don't have a strong preference

web-marketplace/src/components/molecules/CreditsAmount/CreditsInput.tsx
```diff
@@ -35,8 +35,8 @@ export const CreditsInput = ({
   }, [creditsAvailable, paymentOption, setValue]);
 
   const onHandleChange = (event: ChangeEvent<HTMLInputElement>) => {
-    handleCreditsAmountChange(event);
     onChange(event);
+    handleCreditsAmountChange(event);
```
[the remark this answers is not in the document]
blushi: I just think this makes more sense to change the actual form value first, then `handleCreditsAmountChange` which does some computation

blushi:

erikalogie:
This PR is looking absolutely awesome @blushi. Amazing work!

One thing I'm wondering is, when I'm buying with crypto, if I have "retire credits now" selected and have an amount in the fields above, if I then select "buy tradable ecocredits" then the amount resets. Does it need to be this way? If not, could we change this in a follow-up issue?

blushi:
> This PR is looking absolutely awesome @blushi. Amazing work!
> 
> One thing I'm wondering is, when I'm buying with crypto, if I have "retire credits now" selected and have an amount in the fields above, if I then select "buy tradable ecocredits" then the amount resets. Does it need to be this way? If not, could we change this in a follow-up issue?

You're right, it doesn't always need to be reseted.
As an improvement we could:
1. keep the amount if all credits are tradable
2. set amount to `min(current amount, amount credits tradable)` if not all credits are tradable, like here: https://www.figma.com/design/BTuUv6QXY4GbliZcXCe8RJ/Fiat-payments?node-id=174-102754&node-type=frame&m=dev?

web-marketplace/src/pages/BuyCredits/BuyCredits.utils.ts
```diff
@@ -0,0 +1,44 @@
...
+import { PaymentOptionsType } from './BuyCredits.types';
+
+type GetFormModelParams = {
+  _: TranslatorType;
+  paymentOption: PaymentOptionsType;
+  retiring: boolean;
+  projectId: string;
+};
+export const getFormModel = ({
+  _,
```
r41ph, line 18: Another way that I think would be more efficient would be that instead of passing around `_` we can use `i18n` from `import { i18n } from '@lingui/core';`

That way we could directly do: ``i18n._(msg`Choose credits`)``

erikalogie:
> > This PR is looking absolutely awesome @blushi. Amazing work!
> > One thing I'm wondering is, when I'm buying with crypto, if I have "retire credits now" selected and have an amount in the fields above, if I then select "buy tradable ecocredits" then the amount resets. Does it need to be this way? If not, could we change this in a follow-up issue?
> 
> You're right, it doesn't always need to be reseted. As an improvement we could:
> 
> 1. keep the amount if all credits are tradable
> 2. set amount to `min(current amount, amount credits tradable)` if not all credits are tradable, like here: https://www.figma.com/design/BTuUv6QXY4GbliZcXCe8RJ/Fiat-payments?node-id=174-102754&node-type=frame&m=dev?

Yes, that would be great. Do you want to open a follow-up task for this? Otherwise everything LGTM

web-marketplace/src/pages/BuyCredits/BuyCredits.utils.ts
blushi: Reading https://lingui.dev/ref/react#uselingui, I believe this might cause some issues when the user switches languages

r41ph: You're right that there could be issues when switching languages. However, what I understand is that it can cause issues when `i18n._()` is used in memoized functions or components. As long as we avoid using it in those contexts it seems that it should work fine.

blushi:
> > > This PR is looking absolutely awesome @blushi. Amazing work!
> > > One thing I'm wondering is, when I'm buying with crypto, if I have "retire credits now" selected and have an amount in the fields above, if I then select "buy tradable ecocredits" then the amount resets. Does it need to be this way? If not, could we change this in a follow-up issue?
> > 
> > 
> > You're right, it doesn't always need to be reseted. As an improvement we could:
> > 
> > 1. keep the amount if all credits are tradable
> > 2. set amount to `min(current amount, amount credits tradable)` if not all credits are tradable, like here: https://www.figma.com/design/BTuUv6QXY4GbliZcXCe8RJ/Fiat-payments?node-id=174-102754&node-type=frame&m=dev?
> 
> Yes, that would be great. Do you want to open a follow-up task for this? Otherwise everything LGTM

Just added it here

blushi:
> > > > This PR is looking absolutely awesome @blushi. Amazing work!
> > > > One thing I'm wondering is, when I'm buying with crypto, if I have "retire credits now" selected and have an amount in the fields above, if I then select "buy tradable ecocredits" then the amount resets. Does it need to be this way? If not, could we change this in a follow-up issue?
> > > 
> > > 
> > > You're right, it doesn't always need to be reseted. As an improvement we could:
> > > 
> > > 1. keep the amount if all credits are tradable
> > > 2. set amount to `min(current amount, amount credits tradable)` if not all credits are tradable, like here: https://www.figma.com/design/BTuUv6QXY4GbliZcXCe8RJ/Fiat-payments?node-id=174-102754&node-type=frame&m=dev?
> > 
> > 
> > Yes, that would be great. Do you want to open a follow-up task for this? Otherwise everything LGTM
> 
> Just added it here

@erikalogie

blushi:
> Can we make the default number of credits when the flow first loads be 1 rather than 0?

it could be that there are less than 1 credit available, ie 0.5

erikalogie:
> > Can we make the default number of credits when the flow first loads be 1 rather than 0?
> 
> it could be that there are less than 1 credit available, ie 0.5

Could we make it 1 or the highest number available under 1?

clevinson:
Nice work!

Few bugs I want to highlight here as I'm going through QA-

## Input forms

On the dollar / credit input forms we allow for several `0`s to be typed such that a number is `0` padded. I think this shouldn't be allowed, and the moment a user enters a non-zero number, all `0` digits should be removed from the input field.  This likely will get addressed by @erikalogie's comment above to have a non-zero default, but I wanted to highlight this issue explicitly incase it doesn't get resolved already by changing to non-zero defaults.

Screenshot 2024-10-02 at 4 39 33 PM

## Fiat Pathway

No bugs!

## Crypto Pathway

When I:
- select a quantity of credits to purchase (with "retired credits now" selected)
- change the purchase option to "Buy tradable ecocredits"

Then:
- the quantity does not update
- the purchase fails with an error on "auto retire" (see screenshot)

Screenshot 2024-10-02 at 4 04 12 PM

Separate from the above issue, when successfully purchasing tradable credits, the success modal at the end says "View retirement certificate", instead of something like "View ecocredits". I know that we haven't implemented the entire buy flow yet, and there are still follow-ups remaining, but just wanted to make sure this particular issue was being tracked.

blushi:
> Separate from the above issue, when successfully purchasing tradable credits, the success modal at the end says "View retirement certificate", instead of something like "View ecocredits". I know that we haven't implemented the entire buy flow yet, and there are still follow-ups remaining, but just wanted to make sure this particular issue was being tracked.

There's indeed a follow up issue for the confirmation step: https://regennetwork.atlassian.net/browse/APP-361
But when looking at the figma designs, I don't see any confirmation screen when buying tradable credits @erikalogie looks like there's always the retirement certificate page on the page of the confirmation modal.

web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.utils.tsx
```diff
@@ -1,37 +1,171 @@
...
+      );
+}
+
+type GetCreditsAmountParams = {
+  value: number;
+  card: boolean;
+  orderedSellOrders: UISellOrderInfo[];
+  creditTypePrecision?: number | null;
+};
+export const getCreditsAmount = ({
```
blushi, line 50: @r41ph just wanted to mention that those functions `getCreditsAmount ` and `getCurrencyAmount` are one of the most important stuff to review in this PR, since it infers the sell orders (by the cheapest) based on the credits or currency amount provided

blushi:
@erikalogie @clevinson could you have another look?

erikalogie:
LGTM

erikalogie:
> > Separate from the above issue, when successfully purchasing tradable credits, the success modal at the end says "View retirement certificate", instead of something like "View ecocredits". I know that we haven't implemented the entire buy flow yet, and there are still follow-ups remaining, but just wanted to make sure this particular issue was being tracked.
> 
> There's indeed a follow up issue for the confirmation step: https://regennetwork.atlassian.net/browse/APP-361 But when looking at the figma designs, I don't see how the confirmation screen is supposed to look like when buying tradable credits @erikalogie looks like there's always the retirement certificate on the background of the confirmation modal.

You are 100% correct. I just made a version here, wdyt? https://www.figma.com/design/BTuUv6QXY4GbliZcXCe8RJ/Fiat-payments?node-id=5075-76841&t=lKYcqtNVA7T8TxHW-1

blushi:
> > > Separate from the above issue, when successfully purchasing tradable credits, the success modal at the end says "View retirement certificate", instead of something like "View ecocredits". I know that we haven't implemented the entire buy flow yet, and there are still follow-ups remaining, but just wanted to make sure this particular issue was being tracked.
> > 
> > 
> > There's indeed a follow up issue for the confirmation step: https://regennetwork.atlassian.net/browse/APP-361 But when looking at the figma designs, I don't see how the confirmation screen is supposed to look like when buying tradable credits @erikalogie looks like there's always the retirement certificate on the background of the confirmation modal.
> 
> You are 100% correct. I just made a version here, wdyt? https://www.figma.com/design/BTuUv6QXY4GbliZcXCe8RJ/Fiat-payments?node-id=5075-76841&t=lKYcqtNVA7T8TxHW-1

Looks good, just left a comment.

web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.utils.tsx
```diff
@@ -1,37 +1,171 @@
...
+
+  for (let i = 0; i < orderedSellOrders.length; i++) {
+    const order = orderedSellOrders[i];
+    const price = getSellOrderPrice({ order, card });
+    const quantity = Number(order.quantity);
+    const orderTotalAmount = quantity * price;
+
+    if (currencyAmountLeft >= orderTotalAmount) {
+      currencyAmountLeft = parseFloat(
+        (currencyAmountLeft - orderTotalAmount).toFixed(6),
```
r41ph, line 69: Should this be `toFixed(creditTypePrecision || 6)` ?

web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.utils.tsx
```diff
@@ -1,37 +1,171 @@
...
+      );
+      currentCreditsAmount += quantity;
+      sellOrders.push(formatFullSellOrder({ order, card, price }));
+      if (currencyAmountLeft === 0) break;
+    } else {
+      currentCreditsAmount += currencyAmountLeft / price;
+      sellOrders.push({
+        sellOrderId: order.id,
+        quantity: (currencyAmountLeft / price).toFixed(
+          creditTypePrecision || 6,
```
r41ph, line 79: As we are using this more than once maybe it can be extracted to `const precision = creditTypePrecision || 6;`

web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.utils.tsx
```diff
@@ -1,37 +1,171 @@
...
+  card,
+  orderedSellOrders,
+  creditTypePrecision,
+}: GetCreditsAmountParams) => {
+  const currentCurrencyAmount = card ? value : denomToMicro(value);
+  let currentCreditsAmount = 0;
+  let currencyAmountLeft = currentCurrencyAmount;
+  const sellOrders = [];
+
+  for (let i = 0; i < orderedSellOrders.length; i++) {
```
r41ph, line 61: We could just do:
```
for (const order of orderedSellOrders) {
```
and remove this: `const order = orderedSellOrders[i];`

web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.utils.tsx
```diff
@@ -1,37 +1,171 @@
...
+    if (currencyAmountLeft >= orderTotalAmount) {
+      currencyAmountLeft = parseFloat(
+        (currencyAmountLeft - orderTotalAmount).toFixed(6),
+      );
+      currentCreditsAmount += quantity;
+      sellOrders.push(formatFullSellOrder({ order, card, price }));
+      if (currencyAmountLeft === 0) break;
+    } else {
+      currentCreditsAmount += currencyAmountLeft / price;
+      sellOrders.push({
```
r41ph, line 76: We could improve readability by extracting this into a function and potentially use it again in `getCurrencyAmount`: 
```
...
sellOrders.push(
  createOrderDetails(
    order,
    currencyAmountLeft,
    price,
    creditTypePrecision,
    card,
  ),
);
...


function createOrderDetails(
  order: UISellOrderInfo,
  currencyAmountLeft: number,
  price: number,
  creditTypePrecision: number | null | undefined,
  card: boolean,
): any {
  return {
    sellOrderId: order.id,
    quantity: (currencyAmountLeft / price).toFixed(creditTypePrecision || 6),
    bidPrice: !card
      ? { amount: String(price), denom: order.askDenom }
      : undefined,
    price: card ? price * 100 : undefined,
  };
}
```

web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.utils.tsx
```diff
@@ -1,37 +1,171 @@
...
+  currentCreditsAmount,
+  card,
+  orderedSellOrders,
+  creditTypePrecision,
+}: GetCurrencyAmountParams) => {
+  let currentCurrencyAmount = 0;
+  let creditsAmountLeft = currentCreditsAmount;
+  const sellOrders = [];
+
+  for (let i = 0; i < orderedSellOrders.length; i++) {
```
r41ph, line 113: Same as with `getCreditAmount()`, we could just use a `for...of`

web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.utils.tsx
```diff
@@ -1,37 +1,171 @@
...
+        (creditsAmountLeft - quantity).toFixed(creditTypePrecision || 6),
+      );
+      currentCurrencyAmount += quantity * price;
+      sellOrders.push(formatFullSellOrder({ order, card, price }));
+
+      if (creditsAmountLeft === 0) break;
+    } else {
+      // Take only remaining credits
+      currentCurrencyAmount += creditsAmountLeft * price;
+      sellOrders.push({
```
r41ph, line 130: We could use the `createOrderDetails()` function I suggested above here too

web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.utils.tsx
r41ph: @blushi this looks great! I've tested it and it seems to work as expected. Just a few minor suggestions.

web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.utils.tsx
blushi: no here we have a currency amount, not a credits amount so `creditTypePrecision` doesn't apply

web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.utils.tsx
blushi: I've updated our existing `formatFullSellOrder` function to take an optional quantity as param
see https://github.com/regen-network/regen-web/pull/2477/commits/374be18423ad537000c3378df13032dec87460c9#diff-62c8723f280ca5d74f2f49f2027c2c5f403f8c8944c66d6b2eb960874b66c717

r41ph (approved):

clevinson:
Testing again, there's a few edge cases that still are buggy in the lastest build.  I'll try to enumerate them here:

1. you can't enter a number smaller than `0.1`. As soon as you enter `0.0` it changes to `0`, and so numbers like `0.05` or `0.09` are not possible to enter by typing them in
2. When entering a number with more than 2 decimal plages (e.g. 0.499) you get an error screen that persists on refresh and will not go away unless you clear session data (see image below)
3. When putting in a dollar amount less than $0.50, the credit card modal fails to display on the following screen (see image below)
4. When trying to type `$1.05`, it won't let me type `1.0`, and once I type `1.0`, it auto reverts to to `1`.
 
Screenshot 2024-10-09 at 7 11 43 PM
---

Screenshot 2024-10-09 at 10 28 20 AM

blushi:
> Testing again, there's a few edge cases that still are buggy in the lastest build. I'll try to enumerate them here:
> 
> 1. you can't enter a number smaller than `0.1`. As soon as you enter `0.0` it changes to `0`, and so numbers like `0.05` or `0.09` are not possible to enter by typing them in
> 2. When entering a number with more than 2 decimal plages (e.g. 0.499) you get an error screen that persists on refresh and will not go away unless you clear session data (see image below)
> 3. When putting in a dollar amount less than $0.50, the credit card modal fails to display on the following screen (see image below)
> 4. When trying to type `$1.05`, it won't let me type `1.0`, and once I type `1.0`, it auto reverts to to `1`.
> 
> When fixing (2) we should probably also ensure that if there's ever an error screen that a user gets on, that we don't freeze on that screen, but rather clear the local storage so when they refresh they will be brought back to the first step of the purchase.
> 

Thanks for testing again, looks like most of this might be due to the recent modification to get rid of the leading 0, I'll have a look.

blushi:
> When entering a number with more than 2 decimal plages (e.g. 0.499) you get an error screen that persists on refresh and will not go away unless you clear session data (see image below)

This is when you enter a currency amount buying with credit card?
Stripe only accepts numbers in the smallest currency unit, which is the cent in dollar so when you enter $0.499, we provide 0.499*100=49.9 cents to Stripe, which doesn't support it.
I guess we should simply limit the allowed number of decimals to 2 when for the currency amount using card. WDYT?

r41ph:
> Testing again, there's a few edge cases that still are buggy in the lastest build. I'll try to enumerate them here:

@blushi @clevinson this might be related to an issue I've just noticed when after clicking on the 'Max' button the currency amount goes to an amount starting with 0.00... leading to an error in the form when trying to go to the Next step. See attached 

https://github.com/user-attachments/assets/f0994f33-a430-4951-a015-6533833298e9

blushi:
Addressed 1. and 4.

blushi:
> > Testing again, there's a few edge cases that still are buggy in the lastest build. I'll try to enumerate them here:
> 
> @blushi @clevinson this might be related to an issue I've just noticed when after clicking on the 'Max' button the currency amount goes to an amount starting with 0.00... leading to an error in the form when trying to go to the Next step. See attached
> 
>  Screen.Recording.mov

The issue is because we get more than 6 decimals, I'll have a look.

blushi:
> > > Testing again, there's a few edge cases that still are buggy in the lastest build. I'll try to enumerate them here:
> > 
> > 
> > @blushi @clevinson this might be related to an issue I've just noticed when after clicking on the 'Max' button the currency amount goes to an amount starting with 0.00... leading to an error in the form when trying to go to the Next step. See attached
> > Screen.Recording.mov
> 
> The issue is because we get more than 6 decimals, I'll have a look.

just pushed a fix: [cde9ef2](https://github.com/regen-network/regen-web/pull/2477/commits/cde9ef2d8d5c54885c9862fb12e46abb64f5b4e4)

erikalogie:
> > 2. When entering a number with more than 2 decimal plages (e.g. 0.499) you get an error screen that persists on refresh and will not go away unless you clear session data (see image below)
> 
> @clevinson This is when you enter a currency amount buying with credit card? Stripe only accepts numbers in the smallest currency unit, which is the cent in dollar so when you enter $0.499, we provide 0.499*100=49.9 cents to Stripe, which doesn't support it. I guess we should simply limit the allowed number of decimals to 2 for the currency amount when using card. WDYT?
> 
> But then what if the user enters an amount of credits that give a currency amount with more than 2 decimals? should we round it up to 2 decimals and update the credits amount accordingly as well? It would be good to display a banner indicating that cc/ @erikalogie

Could we force that the in the credit amount we also don't allow for more than two decimals on the fiat purchase screen only? cc @clevinson wdyt

> 
> > 3. When putting in a dollar amount less than $0.50, the credit card modal fails to display on the following screen (see image below)
> 
> This seems to be a limitation from stripe as well, see https://docs.stripe.com/currencies#minimum-and-maximum-charge-amounts. I would suggest the same solution as for 2.
> 
> If that's alright, I'd rather address that in a follow-up issue, because we have some other issues that need this PR so we've just based our feature branch of this branch but this can lead to conflicts when rebasing, which are time consuming to resolve.

If Cory doesn't reply quickly, this seems fine to me

blushi:
> > > 2. When entering a number with more than 2 decimal plages (e.g. 0.499) you get an error screen that persists on refresh and will not go away unless you clear session data (see image below)
> > 
> > 
> > @clevinson This is when you enter a currency amount buying with credit card? Stripe only accepts numbers in the smallest currency unit, which is the cent in dollar so when you enter $0.499, we provide 0.499*100=49.9 cents to Stripe, which doesn't support it. I guess we should simply limit the allowed number of decimals to 2 for the currency amount when using card. WDYT?
> > But then what if the user enters an amount of credits that give a currency amount with more than 2 decimals? should we round it up to 2 decimals and update the credits amount accordingly as well? It would be good to display a banner indicating that cc/ @erikalogie
> 
> Could we force that in the credit amount we also don't allow for more than two decimals on the fiat purchase screen only? cc @clevinson wdyt
> 

I'm not sure if it will ever be the case, but having a credit amount with 2 decimals max doesn't necessarily mean the resulting price will also have max 2 decimals, ie if you want to buy 0.01 credit at $0.01/credit then you would end up with $0.0001 to pay, which is not supported.

> > > 3. When putting in a dollar amount less than $0.50, the credit card modal fails to display on the following screen (see image below)
> > 
> > 
> > This seems to be a limitation from stripe as well, see https://docs.stripe.com/currencies#minimum-and-maximum-charge-amounts. I would suggest the same solution as for 2.
> > If that's alright, I'd rather address that in a follow-up issue, because we have some other issues that need this PR so we've just based our feature branch of this branch but this can lead to conflicts when rebasing, which are time consuming to resolve.
> 
> If Cory doesn't reply quickly, this seems fine to me

erikalogie:
> > > > 2. When entering a number with more than 2 decimal plages (e.g. 0.499) you get an error screen that persists on refresh and will not go away unless you clear session data (see image below)
> > > 
> > > 
> > > @clevinson This is when you enter a currency amount buying with credit card? Stripe only accepts numbers in the smallest currency unit, which is the cent in dollar so when you enter $0.499, we provide 0.499*100=49.9 cents to Stripe, which doesn't support it. I guess we should simply limit the allowed number of decimals to 2 for the currency amount when using card. WDYT?
> > > But then what if the user enters an amount of credits that give a currency amount with more than 2 decimals? should we round it up to 2 decimals and update the credits amount accordingly as well? It would be good to display a banner indicating that cc/ @erikalogie
> > 
> > 
> > Could we force that in the credit amount we also don't allow for more than two decimals on the fiat purchase screen only? cc @clevinson wdyt
> 
> @erikalogie I'm not sure if it will ever be the case, but having a credit amount with 2 decimals max doesn't necessarily mean the resulting price will also have max 2 decimals, ie if you want to buy 0.01 credit at $0.01/credit then you would end up with $0.0001 to pay, which is not supported.

You are right, that totally doesn't solve the problem. Here is a prototype that I made (we can have the updating happen when the user leaves the field, as I'm showing in the prototype, or maybe more simple is as soon as they enter an invalid number): https://www.loom.com/share/c55768f886c94be694c7a707dbeec987

Here it is in Figma: https://www.figma.com/design/BTuUv6QXY4GbliZcXCe8RJ/Fiat-payments?node-id=5311-77828&t=7AFABCQDK2t8vxmH-1 

@clevinson I sent these to Marie on Slack and this is also what she had in mind. Will you take a look and let us know if this also makes sense to you as a way to handle this?

clevinson:
Can't we have for USD / fiat currencies always round up to the nearest penny in real time, directly in the input field?

Having the red banner at the top makes it seem like the user did something wrong, so I'd like to avoid that if possible.

erikalogie:
> Can't we have for USD / fiat currencies always round up to the nearest penny in real time, directly in the input field?
> 
> Having the red banner at the top makes it seem like the user did something wrong, so I'd like to avoid that if possible.

Sure, we could try that and see how it feels.

erikalogie:
> > Can't we have for USD / fiat currencies always round up to the nearest penny in real time, directly in the input field?
> > Having the red banner at the top makes it seem like the user did something wrong, so I'd like to avoid that if possible.
> 
> Sure, we could try that and see how it feels.

In this case it would look like this:

https://github.com/user-attachments/assets/ac2d9c1a-20b8-40ce-9571-eab45849da03

erikalogie:
I think you are already fixing this @blushi but just to make sure:
Screen Shot 2024-10-10 at 2 57 24 PM

blushi:
> I think you are already fixing this @blushi but just to make sure: Screen Shot 2024-10-10 at 2 57 24 PM

This should be fixed now @erikalogie 

I've created a follow-up issue for handling limitations related to Stripe: https://regennetwork.atlassian.net/browse/APP-396 @clevinson

erikalogie:
> > I think you are already fixing this @blushi but just to make sure: Screen Shot 2024-10-10 at 2 57 24 PM
> 
> This should be fixed now @erikalogie
> 
> I've created a follow-up issue for handling limitations related to Stripe for the USD amount: https://regennetwork.atlassian.net/browse/APP-396 @clevinson

This appears to be working well. I wonder if we should merge this and solve any further bugs in follow up issues.

blushi:
> > > I think you are already fixing this @blushi but just to make sure: Screen Shot 2024-10-10 at 2 57 24 PM
> > 
> > 
> > This should be fixed now @erikalogie
> > I've created a follow-up issue for handling limitations related to Stripe for the USD amount: https://regennetwork.atlassian.net/browse/APP-396 @clevinson
> 
> This appears to be working well. I wonder if we should merge this and solve any further bugs in follow up issues.

Yeah I believe we should merge this.

PR #2501, opened 2024-10-10
feat: APP-368 add user balance error in buy credits form

r41ph:
## Description

https://regennetwork.atlassian.net/browse/APP-368

### Author Checklist

I have...

### How to test

1.

### Reviewers Checklist

I have...

- [ ] reviewed tests


r41ph:
@blushi 

Relevant files:
web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.stories.tsx
web-marketplace/src/components/molecules/CreditsAmount/CurrencyInput.tsx
web-marketplace/src/components/organisms/BuyCreditsModal/BuyCreditsModal.tsx
web-marketplace/src/components/organisms/BuyCreditsModal/hooks/useFetchUserBalance.tsx
web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.constants.tsx
web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.schema.tsx
web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.tsx

r41ph:
> @r41ph you can also simply have the base branch for this PR be feat-APP-204-buy-credits so the diff is more readable (at least until my PR gets merged)

Got it, thanks! I've changed the base now

web-marketplace/src/components/organisms/BuyCreditsModal/hooks/useFetchUserBalance.tsx
```diff
@@ -9,12 +9,16 @@ import { UISellOrderInfo } from 'pages/Projects/AllProjects/AllProjects.types';
 
 interface Params {
   selectedSellOrder?: UISellOrderInfo;
+  askDenom?: string;
```
blushi, line 12: what about refactoring this to only take `askDenom` as param?

web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.tsx
```diff
@@ -99,11 +102,15 @@ export function ChooseCreditsForm({
 
   const [spendingCap, setSpendingCap] = useState(0);
   const [creditsAvailable, setCreditsAvailable] = useState(0);
+  // A big number here avoids the NOT_ENOUGH_BALANCE error on rendering
+  // the form until the actual user balance is fetched
+  const [userBalance, setUserBalance] = useState(10101010101010101);
```
blushi, line 107: this is a bit hacky, what about displaying a loading spinner instead of the form while we fetch the balance?

web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.constants.tsx
```diff
@@ -3,3 +3,4 @@ import { msg } from '@lingui/macro';
 export const MAX_AMOUNT = msg`Amount cannot exceed`;
 export const MAX_CREDITS = msg`Credits cannot exceed`;
 export const POSITIVE_NUMBER = msg`Must be positive`;
+export const NOT_ENOUGH_BALANCE = msg`You don’t have enough`;
```
blushi, line 6: aren't we supposed to display `You don’t have enough ${displayDenom}`?

blushi (changes requested):

web-marketplace/src/components/organisms/BuyCreditsModal/hooks/useFetchUserBalance.tsx
blushi: it's only used once, so it would be really a small refactor:

`const userBalance = useFetchUserBalance({ askDenom: selectedSellOrder?.askDenom });`

web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.constants.tsx
r41ph: yes, that's what we do here: https://github.com/regen-network/regen-web/pull/2501/files#diff-dc24cb13d9ae4f19768619f0203bc0cbe44374ee901e3edb0bea0e4d6114a8e9R157

blushi: oh I see, that assumes that all errors for this field will have the displayDenom at the end of the message.
I guess that works for now but that's not very reusable and also it gives less context for translating the content which could result in wrong translations.

web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.tsx
r41ph: Since the data we are fetching is related to validation for an error users may never encounter, I thought it'd be better to not block the UI from rendering, so users can interact with the form straight away. But yes, a loading animation is a more conventional solution.

web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.tsx
```diff
@@ -144,11 +161,14 @@ export function ChooseCreditsForm({
   const handlePaymentOptions = useCallback(
     (option: string) => {
       setPaymentOption(option as PaymentOptionsType);
-      form.setValue(CREDIT_VINTAGE_OPTIONS, []);
-      form.setValue(
-        CURRENCY,
-        option === PAYMENT_OPTIONS.CARD ? cardCurrency : defaultCryptoCurrency,
-      );
+      form.reset({
```
blushi, line 164: why do we need to do this?

web-marketplace/src/pages/BuyCredits/BuyCredits.tsx
```diff
@@ -39,7 +39,7 @@ export const BuyCredits = () => {
   } = useGetProject();
 
   const [paymentOption, setPaymentOption] = useState<PaymentOptionsType>(
-    PAYMENT_OPTIONS.CRYPTO,
+    PAYMENT_OPTIONS.CARD,
```
blushi, line 42: why this change? this results in the crypto option not being selected by default if there's no fiat sell orders:

image

the card option is greyed out but we still see the related form and the user needs to click on "buy with crypto" which feels a bit weird

blushi (changes requested):
I had 1 credit selected buying in REGEN (my account had enough REGENs so no error), but then if I select another currency where my balance doesn't exist, I do not get any error

web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.tsx
r41ph: If we don't reset the form, then the form validation, now depending on the `paymentOption` param, doesn't get updated with the newly selected option handled by `handlePaymentOptions`.

blushi: if you just want to trigger validation you could use `shouldValidate` or `trigger`

r41ph: Is not that we need to trigger validation, we want the validation scheme to get updated with the new `paymentOption` value. If I don't use `reset` then I see the `userBalance` error in the USD form. Tried with `shouldValidate` and `trigger` but didn't work

web-marketplace/src/pages/BuyCredits/BuyCredits.tsx
r41ph: The reason behind this change was that the user balance validation was getting triggered in the fiat form because the payment method was crypto.. but I'm unable to reproduce this bug anymore so reverting back to card.

Although, arguably, the default option should be card, and if there is no `cardSellOrders` we should update the `paymentOption` to crypto  in a `useEffect`, wdyt?

blushi: it's done the other way around right now, and switching to card in a use effect if there are some cardSellOrders 
I would prefer to keep it this way just to avoid any extra work

r41ph:

erikalogie:
LGTM - Not important now while we are in a time crunch, but I'll make a follow-up issue that changes the "next" button to say "insufficient balance" as in the comps

blushi (commented):
Looks good but just a small edge case I've found.
If I'm on "buy with crypto" and get the error "You don't have enough ..." then switch to "buy with card", the error remains with USD, it should disappear

r41ph:
Ok that's interesting.. This is the error I previously fixed by using `form.reset()` and couldn't reproduce yesterday after using `trigger` to validate the `userBalance`. I'll see if I can find a solution that doesn't require using `form.reset()`

web-marketplace/src/components/organisms/BuyCreditsModal/hooks/useFetchUserBalance.tsx
```diff
@@ -5,16 +5,15 @@ import { useLedger } from 'ledger';
 import { getAllBalancesQuery } from 'lib/queries/react-query/cosmos/bank/getAllBalancesQuery/getAllBalancesQuery';
```
blushi, line 5: we could actually move this hook to the new  /pages/BuyCredits/hooks and delete BuyCreditsModal entirely since it's not used anymore

web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.tsx
```diff
@@ -163,6 +168,19 @@ export function ChooseCreditsForm({
     name: CURRENCY_AMOUNT,
   });
 
+  const { isLoading, userBalance: microUserBalance } = useFetchUserBalance(
```
blushi, line 171: this breaks https://deploy-preview-2501--regen-storybook.netlify.app/?path=/story/marketplace-organisms-choosecreditsform--choose-credits
maybe we could just pass it as a prop?

blushi: we might need this logic for updating the credits amount from the order summary card so it would make sense to move it one level up

@erikalogie what should we display in case the user enters a amount of credits from the order summary card that results in an amount to pay that exceeds user balance?

r41ph:
@blushi I have we moved the hook to `/pages/BuyCredits/hooks` and deleted `BuyCreditsModal`. Also have moved up to the parent component the `userBalance` logic so we can address the order summary errors when @erikalogie replies to your question - probably better to sort out the order summary errors in a different issue?

Also, I have removed the hook `useBuySellOrderSubmit` that was breaking the build and wasn't being used anywhere since we removed `BuyCreditsModal` - let me know if there's any reason why we should keep it and I add it back.

web-marketplace/src/pages/BuyCredits/hooks/useFetchUserBalance.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { getAllBalancesQuery } from 'lib/queries/react-query/cosmos/bank/getAllBalancesQuery/getAllBalancesQuery';
+import { useWallet } from 'lib/wallet/wallet';
+
+export const useFetchUserBalance = (
+  askDenom?: string,
+): {
+  isLoading: boolean;
+  userBalance: number;
+  microUserBalance: number;
+} => {
```
blushi, line 15: FYI it's fine to not define fn return values when we don't need to explicitly define them as it's the case here

web-marketplace/src/pages/BuyCredits/BuyCredits.Form.tsx
```diff
@@ -117,7 +121,7 @@ export const BuyCreditsForm = ({
   const [paymentOptionCryptoClicked, setPaymentOptionCryptoClicked] = useAtom(
     paymentOptionCryptoClickedAtom,
   );
-
+  const [currencyCryptoAtom] = useAtom(buyCreditsCryptoCurrencyAtom);
```
blushi, line 124: why do we need this in addition to `data?.[CURRENCY]`?

blushi (commented):
could you make sure to run i18n:extract again since you've removed some files which are still referenced in the po files?

web-marketplace/src/pages/BuyCredits/BuyCredits.Form.tsx
r41ph: when going to step 2 using USD and then navigate back to step 1, the currency in `data?.[CURRENCY]` is not updated, even when changing the payment type. This change is to keep track of payment type updates so we can fetch the right `userBalance` - otherwise the balance is always 0 in this scenario.

blushi: I see, this is because `data?.[CURRENCY]` only changes as soon as the user submits step 1
But I find it confusing to have to keep track of this twice in slightly different ways
Maybe we should keep getting the userBalance on the ChooseCreditsForm using `currency` from `useWatch` for this specific case, just like we already instantiate different `filteredCryptoSellOrders` and `orderedSellOrders`

web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.tsx
erikalogie: See this flow starting here: https://www.figma.com/design/BTuUv6QXY4GbliZcXCe8RJ/Fiat-payments?node-id=5314-94485&t=tyf73iuZkQA0S9cW-1

erikalogie: @r41ph @blushi I updated the flow for when it exceeds the user's balance: https://www.figma.com/design/BTuUv6QXY4GbliZcXCe8RJ/Fiat-payments?node-id=5843-92226&t=tyf73iuZkQA0S9cW-1

erikalogie:
@r41ph I would not expect the error in the order summary card to appear unless you click the edit symbol. I'm additionally going to make a follow up task so that you cannot edit the amount from the first step of the flow.

erikalogie:
@r41ph I've also added this task: https://regennetwork.atlassian.net/browse/APP-436

blushi:
> @r41ph I would not expect the error in the order summary card to appear unless you click the edit symbol.

agreed

web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.tsx
```diff
@@ -73,230 +78,268 @@ export type Props = {
...
+          [CURRENCY]:
+            option === PAYMENT_OPTIONS.CARD
+              ? cardCurrency
+              : defaultCryptoCurrency,
+        });
+        const currency =
+          option !== PAYMENT_OPTIONS.CARD
+            ? { askDenom: REGEN_DENOM, askBaseDenom: REGEN_DENOM }
+            : { askDenom: USD_DENOM, askBaseDenom: USD_DENOM };
+        updateMultiStepCurrency(handleSave, data, currency, activeStep);
```
blushi, line 227: why are we setting different values for currency here?
it could even be that REGEN is not available as askDenom

web-marketplace/src/pages/BuyCredits/BuyCredits.tsx
```diff
@@ -36,9 +36,7 @@ export const BuyCredits = () => {
 
   useNavigateToSlug(slug, '/buy');
 
-  const [paymentOption, setPaymentOption] = useState<PaymentOptionsType>(
```
r41ph, line 39: Switched from `useState` to `useAtom` to avoid prop drilling and simplify `paymentOption` state management across nested components in the BuyCredits flow.

r41ph:
> Not a huge issue, but now when switching from crypto to credit card payment, we see for a few ms the USD icon next to MAX button.

This should be fixed now.

r41ph:
@erikalogie have another look please - see updated description.

blushi:
If I try to buy credits from C01-010 (no fiat sell orders available), for some reason, I always get "you don't have enough REGEN" but I do have enough

image

erikalogie:
LGTM! Could we just stack the "update" and "cancel" links on very small mobile screens?
Monosnap Regen Marketplace 2024-10-31 13-21-09

r41ph:
> If I try to buy credits from C01-010 (no fiat sell orders available), for some reason, I always get "you don't have enough REGEN" but I do have enough

@blushi  `data.currency` in `BuyCredits.Form` was `undefined` when there's no fiat sell orders - I have assigned the default crypto currency to `data.currency` when there is no card sell orders.
It'd be great if you could have a close look at the code please. The rebase with the buy success branch was a bit more complicated than expected https://github.com/regen-network/regen-web/pull/2519


> Could we just stack the "update" and "cancel" links on very small mobile screens?

@erikalogie I have stack the buttons and also increased the font size to make the click (touch) a bit easier. Let me know if you'd rather the font to be smaller. Have you considered keeping the Order Summary Card the same on mobile as on desktop screens? I think that could work well too and there'd be space for the buttons without stacking them.

blushi:
Now if I go to https://deploy-preview-2501--regen-marketplace.netlify.app/project/C01-010/buy with a visiting or web2 user, I'm not redirected to the project page. This is how it should behave and it is already the case on dev for projects without fiat credits accessed by non web3 users.

web-marketplace/src/lib/i18n/locales/es.po
```diff
@@ -1,1122 +1,1093 @@
 msgid ""
```
blushi, line 1: please be careful we've translated ES content with AI so running i18n:extract overwrites it
we still need to discuss how we should handle such updates and conflicts but in the mean time, make sure to run also i18n:translate

blushi:
If I go through part of the crypto buy flow for a project then navigate away/close tab and come back to it, then to the first step, I get stuck on this, I need to update the values for the form to be validated:

image

web-marketplace/src/components/molecules/CreditsAmount/CurrencyInput.tsx
```diff
@@ -59,19 +66,25 @@ export const CurrencyInput = ({
...
           ? { askDenom: USD_DENOM, askBaseDenom: USD_DENOM }
           : {
               askDenom,
               askBaseDenom: cryptoCurrencies.filter(
                 cur => cur.askDenom === askDenom,
               )?.[0].askBaseDenom,
-            },
+            };
+      setValue(CURRENCY, currency);
+      updateMultiStepCurrencyAndPaymentOption(
```
blushi, line 79: so now we use both the multi step atom along with some jotai atoms?

erikalogie:
> @erikalogie I have stack the buttons and also increased the font size to make the click (touch) a bit easier. Let me know if you'd rather the font to be smaller. Have you considered keeping the Order Summary Card the same on mobile as on desktop screens? I think that could work well too and there'd be space for the buttons without stacking them.

I think this looks good! Font a bit smaller sounds good too.

Yeah I considered using the desktop Order Summary card but I thought that it was too tall vertically and the main content that the user needs to fill out would be pushed down too far. But we can certainly consider it once we do some more user testing of this feature.

blushi:
> If I go through part of the crypto buy flow for a project then navigate away/close tab and come back to it, then to the first step, I get stuck on this, I need to update the values for the form to be validated:
> 
> image

@r41ph it looks that hasn't been addressed

web-marketplace/src/components/organisms/ChooseCreditsForm/ChooseCreditsForm.tsx
```diff
@@ -73,235 +89,291 @@ export type Props = {
...
-  useEffect(() => {
-    // If there are some sell orders available for fiat purchase, default to 'card' option
-    if (cardSellOrders.length > 0 && !initialPaymentOption) {
-      setPaymentOption(PAYMENT_OPTIONS.CARD);
-      form.setValue(CREDIT_VINTAGE_OPTIONS, []);
-      form.setValue(CURRENCY, cardCurrency);
-    }
-  }, [cardSellOrders.length, initialPaymentOption]); // just run this once
+    useEffect(() => {
+      if (spendingCap === null) {
```
blushi, line 245: I don't believe this is necessarily needed, `setSpendingCap` is already set in `CreditsAmount`
just checking `spendingCap === null` in `createChooseCreditsFormSchema` seems to fix the issue above without this

web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.tsx
```diff
@@ -166,7 +169,9 @@ export const CreditsAmount = ({
   useEffect(() => {
     if (maxCreditsSelected) {
       setValue(CREDITS_AMOUNT, creditsAvailable, { shouldValidate: true });
-      setValue(CURRENCY_AMOUNT, spendingCap, { shouldValidate: true });
+      setValue(CURRENCY_AMOUNT, spendingCap as number, {
```
blushi, line 172: maybe we could just do `if (maxCreditsSelected && spendingCap)` instead of type assertion here?

web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.tsx
```diff
@@ -235,12 +244,10 @@ export const CreditsAmount = ({
         baseDenom={currency.askBaseDenom}
         creditsAvailable={creditsAvailable}
         setMaxCreditsSelected={setMaxCreditsSelected}
-        paymentOption={paymentOption}
       />
       <div className="flex justify-between min-w-full flex-wrap sm:flex-nowrap gap-10 sm:gap-0 items-start">
         <CurrencyInput
-          maxCurrencyAmount={spendingCap}
-          paymentOption={paymentOption}
+          maxCurrencyAmount={spendingCap as number}
```
blushi, line 250: same here, we could use 

```
{spendingCap && (
          <CurrencyInput ...
)}
``` instead of type assertion

blushi (commented):
https://deploy-preview-2501--regen-storybook.netlify.app/?path=/story/marketplace-molecules-creditsamount--credits-amount-card

This story is not really usable now since it shows 0 credits available

web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.tsx
r41ph: In this case should be alright because I don't think `spendingCap` is ever going to be anything other than a number at the time this is executed. That said, it's probably more robust to do `if (maxCreditsSelected && spendingCap)` as you suggest, thanks!

r41ph:
> This story is not really usable now since it shows 0 credits available

I'm not sure if it's worth keeping this story as it's already part of the ChooseCredits one, what do you think?

blushi:
> > This story is not really usable now since it shows 0 credits available
> 
> I'm not sure if it's worth keeping this story as it's already part of the ChooseCredits one, what do you think?

Yeah I guess it's fine to remove it

blushi (approved):
Pre-approving, pending story to remove

PR #2514, opened 2024-10-24
fix: APP-404 update field and trigger validation instead of using onChange

r41ph:
## Description

https://regennetwork.atlassian.net/browse/APP-404

### Author Checklist

I have...

### How to test

1. Go to the Choose Credits step in the Buy Credits flow and verify that the currency input field doesn't throw an error when changing the value
2. https://deploy-preview-2514--regen-marketplace.netlify.app/project/mai-ndombe-4/buy

### Reviewers Checklist

I have...

- [ ] reviewed tests

### DIFF
--- a/web-marketplace/src/components/molecules/CreditsAmount/CurrencyInput.tsx
+++ b/web-marketplace/src/components/molecules/CreditsAmount/CurrencyInput.tsx
@@ -39,6 +39,8 @@ export const CurrencyInput = ({
   } = useFormContext<ChooseCreditsFormSchemaType>();
   const { _ } = useLingui();
 
+  const { onChange } = register(CURRENCY_AMOUNT);
+
   const currency = useWatch({
     control,
     name: CURRENCY,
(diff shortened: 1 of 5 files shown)


r41ph:

erikalogie:
LGTM

web-marketplace/src/components/molecules/CreditsAmount/CurrencyInput.tsx
```diff
@@ -39,22 +39,18 @@ export const CurrencyInput = ({
...
   const currency = useWatch({
     control,
     name: CURRENCY,
   });
 
   const handleOnChange = useCallback(
     (event: ChangeEvent<HTMLInputElement>) => {
-      // Remove zeros in non decimal values and update the value
-      const value = event.target.value;
-      if (!value.includes('.')) setValue(CURRENCY_AMOUNT, Number(value));
```
blushi, line 53: removing this introduces another bug, you can't type a number with 2 decimals if the first decimal is 0, eg 21,05

r41ph: Reverting doesn't seem to work either. There seems to be an error that may be related to stripe when introducing numbers 2.01. Any idea what the thrown error is about @blushi . Also the user gets stuck in the error screen because the amount introduced is in the Local Storage. 

https://github.com/user-attachments/assets/5138d4d8-3f61-43f4-9af6-71c40968c975

r41ph: I think that should be a separated issue btw. If we are happy with the fix for the current issue we may want to merge this so other features don't have to deal with not being able to use the currency input.

r41ph: ok, thanks -  I have added a possible solution to this too

blushi (approved):

PR #2516, opened 2024-10-26
feat: APP-326 project page tebu factors UI

wgardiner:
## Description

Closes: [APP-326](https://regennetwork.atlassian.net/browse/APP-326?atlOrigin=eyJpIjoiYzIxYWEyZWQ0M2FjNGYzM2ExNmM4NGQ0MjhmNDg5NjkiLCJwIjoiaiJ9)

---

### Author Checklist

I have...

### How to test

1. Use storybook to review [Terrasos/Cards/TebuCard entries ](https://deploy-preview-2516--regen-storybook.netlify.app/?path=/story/terrasos-cards-tebucard--area-actions-body-story)

### Reviewers Checklist

I have...

- [ ] reviewed tests

[APP-326]: https://regennetwork.atlassian.net/browse/APP-326?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


blushi:
@erikalogie could you have a look at this? see testing instructions

flagrede (approved):
Looks good!

I just noticed that the tooltip lacked spacing in low resolutions.

image

wgardiner:
> These look great!
> 
> A few small things: The numerical factors that are in the comps are from their protocol. So for example if an ecosystem is "near threatened" then the factor is 0.12, always. Or if the "ecological connectivity" is moderate, the factor is 0.16, always.
> 
> The tooltip icon should be grey for Terrasos instead of green outline.
> 
> All the tooltip copy is [here](https://www.figma.com/design/brkxGV5qNOkZUp0YQl1cxO/Terrasos-Phase-1?node-id=1312-149662&t=znCeKI1qY3Xr9Auh-1) and [here](https://www.figma.com/design/brkxGV5qNOkZUp0YQl1cxO/Terrasos-Phase-1?node-id=1312-149661&t=znCeKI1qY3Xr9Auh-1). The copy varies slightly with the version of the protocol.
> 
> For "extinct" and "extinct in the wild", can we make the color black and red, respectively? I don't think that we will ever use these categories as they don't appear in [their protocol](https://cdn.prod.website-files.com/65ae9262fbf418e225c9e4f0/67059ffc20367f1762a0f659_240909%20ENG%20-%20Protocol%20Biodiversity%20Units%20V4%20-%20Terrasos_Final.docx.pdf), (see page 29-31), but maybe it is good to have anyway since you did it already. I have added these two plus "data deficient" and "not evaluated" to the Figma comps with definitions.

Thanks for the additional info on the factors. Since this is just a UI PR and we're handling internationalization, the factor and description fields are just example slots for now. As I integrate it I'll add the actual data. In modeling the actual data for the project page I tried to just store a single value that we could derive the rest of the UI from, so for example I only store the number for ecological connectivity index and then we can derive the factor and the text from it. I may end up having to revise the data model.

The current tooltip color is dependent on the active client, so it'll change once it's embedded in the Terrasos marketplace and we can tweak it then. Since it's just in Storybook right now, it's falling back to the default Regen color.

Thanks for the tooltip copy, that will be useful in my next PR, integrating it.

I just updated the IUCN Threat indicator colors as you mentioned. Thanks for adding those to the comp. Yeah they're not in their data, but I figured I might as well include them in case we want to reuse this.

PR #2544, opened 2024-11-13
feat: APP-396 two decimal digits for usd amount

r41ph:
## Description

https://regennetwork.atlassian.net/browse/APP-396

### Author Checklist

I have...

### How to test

1. https://deploy-preview-2544--regen-marketplace.netlify.app/project/mai-ndombe-4/buy
2. Verify that the currency input when set to USD only accepts two decimals.
3. Verify that an error is shown when the amount of USD is less than 0.50

### Reviewers Checklist

I have...

- [ ] reviewed tests

### DIFF
--- a/web-components/src/components/stepper/index.tsx
+++ b/web-components/src/components/stepper/index.tsx
@@ -132,7 +132,6 @@ const RegenStepper = ({
                   styles.alternativeLabel,
                   onStepClick && styles.hover,
                   classes?.stepLabel,
-                  'hover:cursor-default',
                 ),
                 iconContainer:
                   classes?.stepIcon ||
--- a/web-components/src/components/stepper/index.tsx
+++ b/web-components/src/components/stepper/index.tsx
@@ -132,6 +132,7 @@ const RegenStepper = ({
                   styles.alternativeLabel,
                   onStepClick && styles.hover,
                   classes?.stepLabel,
+                  'hover:cursor-default',
                 ),
                 iconContainer:
                   classes?.stepIcon ||
--- a/web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.utils.tsx
+++ b/web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.utils.tsx
@@ -29,10 +29,11 @@ export function getSpendingCap(
   cardSellOrders: Array<CardSellOrder>,
 ) {
   return paymentOption === PAYMENT_OPTIONS.CARD
-    ? cardSellOrders.reduce(
-        (prev, cur) => prev + Number(cur.quantity) * cur.usdPrice,
-        0,
-      )
+    ? cardSellOrders.reduce((prev, cur) => {
+        return formatCurrencyAmountTwoDecimals(
+          prev + Number(cur.quantity) * cur.usdPrice,
+        );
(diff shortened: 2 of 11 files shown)


r41ph:
@erikalogie one thing to consider is if we want to round up the currency amount when in USD because when we cut the decimals o two we may end up charging less than the actual credits price - let me know what you think

blushi:
> @erikalogie one thing to consider is if we want to round up the currency amount when in USD because when we cut the decimals o two we may end up charging less than the actual credits price - let me know what you think

@erikalogie @S4mmyb any thoughts on this?
Imagine the exact amount is $10.897, should we charge 10.89 or 10.90?

erikalogie:
> > @erikalogie one thing to consider is if we want to round up the currency amount when in USD because when we cut the decimals o two we may end up charging less than the actual credits price - let me know what you think
> 
> @erikalogie @S4mmyb any thoughts on this? Imagine the exact amount is $10.897, should we charge 10.89 or 10.90?

I think we should round up to $10.90 in this case

blushi:
> > > @erikalogie one thing to consider is if we want to round up the currency amount when in USD because when we cut the decimals o two we may end up charging less than the actual credits price - let me know what you think
> > 
> > 
> > @erikalogie @S4mmyb any thoughts on this? Imagine the exact amount is $10.897, should we charge 10.89 or 10.90?
> 
> I think we should round.

> I think we should round up to $10.90 in this case

Ok so if it were $10.894, we would charge 10.89.

S4mmyb:
> > > > @erikalogie one thing to consider is if we want to round up the currency amount when in USD because when we cut the decimals o two we may end up charging less than the actual credits price - let me know what you think
> > > 
> > > 
> > > @erikalogie @S4mmyb any thoughts on this? Imagine the exact amount is $10.897, should we charge 10.89 or 10.90?
> > 
> > 
> > I think we should round.
> 
> > I think we should round up to $10.90 in this case
> 
> Ok so if it were $10.894, we would charge 10.89.

Yep! <5 we round down, >=5 round up for that third decimal place.

r41ph:
@blushi @erikalogie @S4mmyb have a look at this please

erikalogie:
I think the translation string is missing for the error:
Screenshot 2024-12-03 at 11 41 28 AM

blushi (commented):
Looks good overall. One thing: If I type leading zeros in the currency input, then it's only transformed into a correct value, eg 0.1 when I type the final 1, while for the credits amount input, as long as I have one leading zero, I can't get others
I think both inputs should have consistent behaviors

image

blushi (approved):
Pre-approving
Pending @erikalogie to have another look and translation to be fixed:

image
this shows up when I enter an amount lower than < 50 cents

erikalogie:
> Pre-approving Pending @erikalogie to have another look and translation to be fixed:
> 
> image this shows up when I enter an amount lower than < 50 cents

Let me know when this is fixed and I'll test again @r41ph

r41ph:
@erikalogie please have look, the translation should be fixed now

blushi:
> @erikalogie please have look, the translation should be fixed now

@r41ph could you run i18n:extract so it translates the missing es translations for "Must be at least"?

PR #2572, opened 2024-12-12
fix APP 407 inputs logic

r41ph:
## Description

https://regennetwork.atlassian.net/browse/APP-407

### Author Checklist

I have...

### How to test

https://deploy-preview-2572--regen-marketplace.netlify.app/project/mai-ndombe-4/buy

1. Check the reported bug and the zero leading functionality mentioned in ticket work as expected

### Reviewers Checklist

I have...

- [ ] reviewed tests

### DIFF
--- a/web-marketplace/src/components/molecules/CreditsAmount/CreditsInput.tsx
+++ b/web-marketplace/src/components/molecules/CreditsAmount/CreditsInput.tsx
@@ -31,7 +31,15 @@ export const CreditsInput = ({
   const handleInput = useCallback(
     (event: ChangeEvent<HTMLInputElement>): void => {
       let value = event.target.value;
+
+      // Check if the value starts with a dot and add a leading zero
+      if (/^\.[0-9]/.test(value)) {
+        setValue(CREDITS_AMOUNT, +`0${value}`, {
+          shouldValidate: true,
+        });
+      }
       if (/^0[0-9]/.test(value)) {
+        // remove extra leading zeros
         value = value.replace(/^0+/, '');
         setValue(CREDITS_AMOUNT, parseFloat(Number(value).toFixed(2)), {
           shouldValidate: true,
@@ -46,6 +54,16 @@ export const CreditsInput = ({
       // If the value is empty, set it to 0
       const value = event.target.value;
       if (value === '') {
+        setValue(CREDITS_AMOUNT, 0, { shouldValidate: true });
+      }
+    },
+    [setValue],
+  );
+  const handleOnFocus = useCallback(
+    (event: FocusEvent<HTMLInputElement>): void => {
+      // If the value is 0, set it to '
+      const value = event.target.value;
+      if (value === '0') {
         setValue(CREDITS_AMOUNT, 0, { shouldValidate: true });
       }
     },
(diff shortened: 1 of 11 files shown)


r41ph:
@erikalogie @blushi see testing instructions

Note: I've tried to mimic the behaviour in the video shared in the Jira task but since we are using an `<input type="number">` there are some limitations on what we can achieve, particularly regarding cursor position.

r41ph:
@erikalogie have a look please. I have updated the code to allow typing two decimals that are zero, but in some scenarios the number may be converted to just `2`, like when changing payment method or returning to step 1. This is because our input field is of `type=number` and JavaScript doesn’t distinguish between `2` and `2.00` because both are treated as the same number.

erikalogie:
> @erikalogie have a look please. I have updated the code to allow typing two decimals that are zero, but in some scenarios the number may be converted to just `2`, like when changing payment method or returning to step 1. This is because our input field is of `type=number` and JavaScript doesn’t distinguish between `2` and `2.00` because both are treated as the same number.

This looks great now!

web-marketplace/src/components/molecules/CreditsAmount/CreditsInput.tsx
```diff
@@ -51,6 +59,16 @@ export const CreditsInput = ({
     },
     [setValue],
   );
+  const handleOnFocus = useCallback(
+    (event: FocusEvent<HTMLInputElement>): void => {
+      // If the value is 0, set it to '
```
blushi, line 64: missing end of comment?

web-marketplace/src/components/molecules/CreditsAmount/CurrencyInput.tsx
```diff
@@ -56,32 +56,50 @@ export const CurrencyInput = ({
...
+      const value = event.target.value;
+      const decimalPart = value.split('.')?.[1];
+      if (
+        decimalPart &&
+        decimalPart.length > 2 &&
+        decimalPart.startsWith('00') &&
+        paymentOption === PAYMENT_OPTIONS.CARD
+      ) {
+        event.target.value = value.slice(0, -1);
+      }
```
blushi, line 69: we could refactor to use a function to get the `if` value, as the same check is also used in `handleInput`

blushi (commented):
Looks good, just a few nits.

Just one small difference I noticed (that was already there btw) between the editable input in the order summary card and the `CreditsInput` and `CurrencyInput` is that within the editable input, I cannot use a comma as a decimal separator which is a convention that is widely used: https://en.wikipedia.org/wiki/Decimal_separator#/media/File:DecimalSeparator.svg, looks like we could update `sanitizeValue` function to account for that, but can be in a separate issue if you want?

web-marketplace/src/components/molecules/CreditsAmount/CreditsInput.tsx
r41ph: I'll removed it

web-marketplace/src/components/molecules/CreditsAmount/CurrencyInput.tsx
r41ph: While they are not exactly the same, refactoring makes sense, thanks!

r41ph:
> Just one small difference I noticed (that was already there btw) between the editable input in the order summary...

After having a look, I've created a new issue for this because it'll require a bit more work than just update `sanitizeValue` because JS doesn't seem to recognise commas as decimal separators. 

https://regennetwork.atlassian.net/browse/APP-540

blushi (approved):

PR #2574, opened 2025-01-08
fix: APP-410 save buy flow on last user interaction

r41ph:
## Description

https://regennetwork.atlassian.net/browse/APP-410

Now when users interact with the forms the information should be stored in the localstorage automatically even when the next button hasn't been clicked yet, and therefore, on navigating away and back to the same page, users should be in the same step and see the same form data as before navigating away.  

---

### Author Checklist

I have...

### How to test

1. https://deploy-preview-2574--regen-marketplace.netlify.app/project/mai-ndombe-4/buy
2. Edit the form and navigate away then come back to the same page and verify the step and the form data is the same as before navigating away. 

### Reviewers Checklist

I have...

- [ ] reviewed tests

### DIFF
--- a/web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.tsx
+++ b/web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.tsx
@@ -12,10 +12,12 @@ import {
   spendingCapAtom,
 } from 'pages/BuyCredits/BuyCredits.atoms';
 import { PAYMENT_OPTIONS } from 'pages/BuyCredits/BuyCredits.constants';
+import { BuyCreditsSchemaTypes } from 'pages/BuyCredits/BuyCredits.types';
 import {
   getCreditsAvailableBannerText,
   getOrderedSellOrders,
 } from 'pages/BuyCredits/BuyCredits.utils';
+import { useMultiStep } from 'components/templates/MultiStepTemplate';
 
 import { findDisplayDenom } from '../DenomLabel/DenomLabel.utils';
 import {
@@ -54,6 +56,11 @@ export const CreditsAmount = ({
   const setWarningBannerTextAtom = useSetAtom(warningBannerTextAtom);
   const [spendingCap, setSpendingCap] = useAtom(spendingCapAtom);
   const paymentOption = useAtomValue(paymentOptionAtom);
+  const {
+    data,
+    handleSave: updateMultiStepData,
+    activeStep,
+  } = useMultiStep<BuyCreditsSchemaTypes>();
 
   const card = useMemo(
     () => paymentOption === PAYMENT_OPTIONS.CARD,
@@ -156,6 +163,16 @@ export const CreditsAmount = ({
         setWarningBannerTextAtom(
           getCreditsAvailableBannerText(_creditsAvailable, displayDenom),
         );
+        // Udates the currency amount in multistep data when payment option is changed
+        updateMultiStepData(
+          {
+            ...data,
+            creditsAmount: _creditsAvailable,
+            currencyAmount: _spendingCap,
+            sellOrders: orderedSellOrders,
+          },
+          activeStep,
+        );
       } else {
         // Else we keep the same amount of credits
         // but we still need to update currency amount and sell orders
@@ -168,8 +185,21 @@ export const CreditsAmount = ({
         });
         setValue(CURRENCY_AMOUNT, currencyAmount);
         setValue(SELL_ORDERS, sellOrders);
+        // Updates the currency amount in multistep data when payment option is changed
+        updateMultiStepData(
+          {
+            ...data,
+            creditsAmount: currentCreditsAmount,
+            currencyAmount: currencyAmount,
+            sellOrders: sellOrders,
+          },
+          activeStep,
+        );
       }
     }
+    // Intentionally omit `updateMultiStepData` and `data` from the dependency array
+    // because including them trigger unnecessary renders.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [
     cardSellOrders,
     filteredCryptoSellOrders,
@@ -183,6 +213,7 @@ export const CreditsAmount = ({
     card,
     setWarningBannerTextAtom,
     _,
+    activeStep,
     displayDenom,
   ]);
 
@@ -193,26 +224,35 @@ export const CreditsAmount = ({
       setValue(CURRENCY_AMOUNT, spendingCap, {
         shouldValidate: true,
       });
-      setValue(
-        SELL_ORDERS,
-        orderedSellOrders.map(order => {
-          const price = getSellOrderPrice({ order, card });
-          return formatSellOrder({ order, card, price });
-        }),
+      const sellOrders = orderedSellOrders.map(order => {
+        const price = getSellOrderPrice({ order, card });
+        return formatSellOrder({ order, card, price });
+      });
+      setValue(SELL_ORDERS, sellOrders);
+      updateMultiStepData(
+        {
+          ...data,
+          creditsAmount: creditsAvailable,
+          currencyAmount: spendingCap,
+          sellOrders: sellOrders,
+        },
+        activeStep,
       );
       setMaxCreditsSelected(false);
     }
   }, [
+    activeStep,
     card,
     creditsAvailable,
+    data,
     maxCreditsSelected,
     orderedSellOrders,
     paymentOption,
     setValue,
     spendingCap,
+    updateMultiStepData,
   ]);
 
-  // Credits amount change
   const handleCreditsAmountChange = useCallback(
     (e: ChangeEvent<HTMLInputElement>) => {
       // Set currency amount according to credits quantity,
@@ -226,11 +266,21 @@ export const CreditsAmount = ({
       });
       setValue(CURRENCY_AMOUNT, currencyAmount, { shouldValidate: true });
       setValue(SELL_ORDERS, sellOrders);
+      updateMultiStepData(
+        {
+          ...data,
+          creditsAmount: currentCreditsAmount,
+          currencyAmount: currencyAmount,
+        },
+        activeStep,
+      );
     },
-    [card, orderedSellOrders, setValue, creditTypePrecision],
+    // Intentionally omit `updateMultiStepData` and `data` from the dependency array
+    // because including them trigger unnecessary renders.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [card, orderedSellOrders, creditTypePrecision, setValue, activeStep],
   );
 
-  // Currency amount change
   const handleCurrencyAmountChange = useCallback(
     (e: ChangeEvent<HTMLInputElement>) => {
       // Set credits quantity according to currency amount,
@@ -242,10 +292,27 @@ export const CreditsAmount = ({
         orderedSellOrders,
         creditTypePrecision,
       });
+
+      updateMultiStepData(
+        {
+          ...data,
+          creditsAmount: currentCreditsAmount,
+          currencyAmount: isNaN(value) ? 0 : value,
+        },
+        activeStep,
+      );
       setValue(CREDITS_AMOUNT, currentCreditsAmount, { shouldValidate: true });
       setValue(SELL_ORDERS, sellOrders);
     },
-    [card, orderedSellOrders, setValue, creditTypePrecision],
+    [
+      card,
+      orderedSellOrders,
+      creditTypePrecision,
+      updateMultiStepData,
+      data,
+      activeStep,
+      setValue,
+    ],
   );
 
   return (
--- a/web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.tsx
+++ b/web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.tsx
@@ -16,6 +16,7 @@ import { BuyCreditsSchemaTypes } from 'pages/BuyCredits/BuyCredits.types';
 import {
   getCreditsAvailableBannerText,
   getOrderedSellOrders,
+  resetCurrencyAndCredits,
 } from 'pages/BuyCredits/BuyCredits.utils';
 import { useMultiStep } from 'components/templates/MultiStepTemplate';
 
@@ -149,49 +150,32 @@ export const CreditsAmount = ({
       // This can happen when the user switches payment option, currency,
       // or to only buy tradable credits,
       // but the amount set is above the amount of newly available credits
-      const currentCreditsAmount = getValues(CREDITS_AMOUNT);
+      let currentCreditsAmount = getValues(CREDITS_AMOUNT);
       if (currentCreditsAmount > _creditsAvailable) {
-        setValue(CREDITS_AMOUNT, _creditsAvailable);
-        setValue(CURRENCY_AMOUNT, _spendingCap);
-        const sellOrders = orderedSellOrders.map(order => {
-          const price = getSellOrderPrice({ order, card });
-          return formatSellOrder({ order, card, price });
-        });
-        setValue(SELL_ORDERS, sellOrders);
         setWarningBannerTextAtom(
           getCreditsAvailableBannerText(_creditsAvailable, displayDenom),
         );
-        // Udates the currency amount in multistep data when payment option is changed
-        updateMultiStepData(
-          {
-            ...data,
-            creditsAmount: _creditsAvailable,
-            currencyAmount: _spendingCap,
-            sellOrders,
-          },
+        resetCurrencyAndCredits(
+          paymentOption,
+          orderedSellOrders,
+          creditTypePrecision,
+          setValue,
+          updateMultiStepData,
+          data,
           activeStep,
+          _creditsAvailable,
         );
       } else {
         // Else we keep the same amount of credits
-        // but we still need to update currency amount and sell orders
-        // (because pricing and sell orders can be different)
-        const { currencyAmount, sellOrders } = getCurrencyAmount({
-          currentCreditsAmount,
-          card,
+        resetCurrencyAndCredits(
+          paymentOption,
           orderedSellOrders,
           creditTypePrecision,
-        });
-        setValue(CURRENCY_AMOUNT, currencyAmount);
-        setValue(SELL_ORDERS, sellOrders);
-        // Updates the currency amount in multistep data when payment option is changed
-        updateMultiStepData(
-          {
-            ...data,
-            creditsAmount: currentCreditsAmount,
-            currencyAmount: currencyAmount,
-            sellOrders: sellOrders,
-          },
+          setValue,
+          updateMultiStepData,
+          data,
           activeStep,
+          currentCreditsAmount,
         );
       }
     }
@@ -269,6 +253,7 @@ export const CreditsAmount = ({
           ...data,
           creditsAmount: currentCreditsAmount,
           currencyAmount: currencyAmount,
+          sellOrders,
         },
         activeStep,
       );
@@ -301,6 +286,7 @@ export const CreditsAmount = ({
           ...data,
           creditsAmount: currentCreditsAmount,
           currencyAmount: isNaN(value) ? 0 : value,
+          sellOrders,
         },
         activeStep,
       );
--- a/web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.tsx
+++ b/web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.tsx
@@ -153,13 +153,11 @@ export const CreditsAmount = ({
       if (currentCreditsAmount > _creditsAvailable) {
         setValue(CREDITS_AMOUNT, _creditsAvailable);
         setValue(CURRENCY_AMOUNT, _spendingCap);
-        setValue(
-          SELL_ORDERS,
-          orderedSellOrders.map(order => {
-            const price = getSellOrderPrice({ order, card });
-            return formatSellOrder({ order, card, price });
-          }),
-        );
+        const sellOrders = orderedSellOrders.map(order => {
+          const price = getSellOrderPrice({ order, card });
+          return formatSellOrder({ order, card, price });
+        });
+        setValue(SELL_ORDERS, sellOrders);
         setWarningBannerTextAtom(
           getCreditsAvailableBannerText(_creditsAvailable, displayDenom),
         );
@@ -169,7 +167,7 @@ export const CreditsAmount = ({
             ...data,
             creditsAmount: _creditsAvailable,
             currencyAmount: _spendingCap,
-            sellOrders: orderedSellOrders,
+            sellOrders,
           },
           activeStep,
         );
--- a/web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.tsx
+++ b/web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.tsx
@@ -273,10 +273,15 @@ export const CreditsAmount = ({
         activeStep,
       );
     },
-    // Intentionally omit `updateMultiStepData` and `data` from the dependency array
-    // because including them trigger unnecessary renders.
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [card, orderedSellOrders, creditTypePrecision, setValue, activeStep],
+    [
+      card,
+      orderedSellOrders,
+      creditTypePrecision,
+      setValue,
+      updateMultiStepData,
+      data,
+      activeStep,
+    ],
   );
 
   const handleCurrencyAmountChange = useCallback(
--- a/web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.tsx
+++ b/web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.tsx
@@ -331,12 +331,16 @@ export const CreditsAmount = ({
             cryptoCurrencies={cryptoCurrencies}
             displayDenom={displayDenom}
             allowedDenoms={allowedDenoms}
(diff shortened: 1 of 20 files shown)


r41ph:

erikalogie:
LGTM

web-marketplace/src/components/organisms/PaymentInfoForm/PaymentInfoForm.tsx
```diff
@@ -89,6 +94,18 @@ export const PaymentInfoForm = ({
     () => paymentOption === PAYMENT_OPTIONS.CARD,
     [paymentOption],
   );
+
+  const handleOnBlur = () => {
+    updateMultiStepData(
+      {
+        ...data,
+        name: form.watch('name'),
+        email: form.watch('email'),
```
blushi, line 103: I believe `form.getValues` https://react-hook-form.com/docs/useform/getvalues would be more appropriate in this case, since we don't need to subscribe to value changes in this function

blushi (changes requested):
Two additional comments:
- We don't save values for the last step to local storage
- When using credit card, if the user leaves the page while being on the last step, when he comes back, we should go back to step 2 (Payment info) because we need the user to re-enter credit card info since this is something we can't save to local storage of course

web-marketplace/src/components/molecules/CreditsAmount/CurrencyInput.tsx
```diff
@@ -97,9 +99,18 @@ export const CurrencyInput = ({
...
         setValue(CURRENCY_AMOUNT, 0, { shouldValidate: true });
+        updateMultiStepData(
+          {
+            ...data,
+            currency,
+            paymentOption,
+            currencyAmount: 0,
+          },
+          activeStep,
+        );
```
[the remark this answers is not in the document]
r41ph: I'm updating the multistep data here to keep it in sync with the input field, don't you think this needs to be done?

web-marketplace/src/components/organisms/PaymentInfoForm/PaymentInfoForm.tsx
r41ph: You are right

web-marketplace/src/components/molecules/CreditsAmount/CreditsAmount.tsx
```diff
@@ -136,6 +143,16 @@ export const CreditsAmount = ({
         setErrorBannerTextAtom(
           getCreditsAvailableBannerText(_creditsAvailable),
         );
+        // Udates the currency amount in multistep data when payment option is changed
+        updateMultiStepData(
+          {
+            ...data,
+            creditsAmount: _creditsAvailable,
+            currencyAmount: _spendingCap,
+            sellOrders: orderedSellOrders,
```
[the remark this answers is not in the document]
r41ph: Good you noticed this, thanks

web-marketplace/src/components/molecules/CreditsAmount/CurrencyInput.tsx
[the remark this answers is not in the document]
blushi: I'm not sure since it seems like this will be updated anyway from `handleCurrencyAmountChange` with the correct values for corresponding sell orders and credits input

r41ph:
@erikalogie can you review the changes mentioned in this comment please https://github.com/regen-network/regen-web/pull/2574#pullrequestreview-2539091513

erikalogie:
LGTM

S4mmyb:
LGTM !

web-marketplace/src/components/organisms/AgreePurchaseForm/AgreePurchaseForm.Retirement.tsx
```diff
@@ -53,6 +61,59 @@ export const Retirement = ({ retiring }: Props) => {
...
+  const retirementReason = useWatch({
+    control: control,
+    name: 'retirementReason',
+  });
+  const postalCode = useWatch({
+    control: control,
+    name: 'postalCode',
+  });
+
+  const { anonymousPurchase, followProject, subscribeNewsletter, agreeErpa } =
```
blushi, line 73: it would be more appropriate to move that `getValues()` into the `useEffect` that uses them, looks like we don't need them at the component top level

this is also going to cause a bug when we enable `anonymousPurchase` checkbox (commented for now) because we use it in the `checked` prop and `getValues` does not subscribe to input change, which will cause the checkbox to be unresponsive, that's why we use `useWatch` instead, which you commented

blushi (commented):
If I set credits or currency amount to 0 then leave the page, now I get this error and I need to reset my buy-credits-... local storage value to make it work again

image

web-marketplace/src/components/organisms/AgreePurchaseForm/AgreePurchaseForm.Retirement.tsx
r41ph: I believe this would only cause a bug if we wouldn't update the `getValues` when we uncomment the `anonymousPurchase` field. However, it'd probably be better to use `useWatch` for all fields. I've made the update, please review.

r41ph:
> If I set credits or currency amount to 0 then leave the page, now I get this error...

I've changed the way we handle `onBlur` when a user leaves the credits and currency inputs. Now if a user leaves the input with a value of `0` or empty `''` I'm resetting the credits to `1` and then setting the corresponding currency amount. Also, I have updated the `stripeOptions` so it doesn't initialise with `0` when `data.[CURRENCY_AMOUNT]` is `undefined`.

@erikalogie @S4mmyb  have a look how the currency and credits fields behave now when a user leaves the inputs with a value of zero or empty, let me know if you want to change anything.

@blushi although the CI shows some failed tests they weren't related with the code updates and a redeploy seems to have been successful: https://app.netlify.com/sites/regen-website/deploys/6788180794db4932a06dce1f

web-marketplace/src/pages/BuyCredits/BuyCredits.Form.tsx
```diff
@@ -278,6 +287,24 @@ export const BuyCreditsForm = ({
...
   );
 
+  const stripeAmount =
+    data?.[CURRENCY_AMOUNT] ||
+    getCurrencyAmount({
+      currentCreditsAmount: 1,
+      card: paymentOption === PAYMENT_OPTIONS.CARD,
+      orderedSellOrders: cardSellOrders,
+      creditTypePrecision: creditTypeData?.creditType?.precision,
+    }).currencyAmount;
```
blushi, line 297: should we put this in a `useMemo` too? `getCurrencyAmount` is a bit computation heavy

blushi (commented):
> I've changed the way we handle `onBlur` when a user leaves the credits and currency inputs. Now if a user leaves the input with a value of `0` or empty `''` I'm resetting the credits to `1` and then setting the corresponding currency amount. 

We should also reset the SELL_ORDERS value in this case. If I set one of the field to 0, then the currency/amount fields do update but not the selected sell orders

Here's how my local storage looks like after that:

image

and then when I go through the end of the flow, I cannot purchase because I'm getting an "invalid amount" error from our server endpoint /marketplace/v1/stripe/create-payment-intent since quantity is `NaN`

erikalogie:
> @erikalogie @S4mmyb have a look how the currency and credits fields behave now when a user leaves the inputs with a value of zero or empty, let me know if you want to change anything.

I think this is fine? @S4mmyb wdyt

blushi (commented):
Looks like there has been a regression with the last few commits because this was working before:
If I go through the credit card flow until last "Retirement" step but leave the page and come back, I'm still at the "Retirement" step while I should be at "Payment info" step to re-enter credit card info

web-marketplace/src/components/molecules/CreditsAmount/CurrencyInput.tsx
```diff
@@ -95,11 +108,26 @@ export const CurrencyInput = ({
...
+        const { currencyAmount, sellOrders } = getCurrencyAmount({
+          currentCreditsAmount: 1,
+          card: paymentOption === PAYMENT_OPTIONS.CARD,
+          orderedSellOrders,
+          creditTypePrecision,
+        });
+        setValue(CURRENCY_AMOUNT, currencyAmount, { shouldValidate: true });
+        setValue(CREDITS_AMOUNT, 1, { shouldValidate: true });
+        setValue(SELL_ORDERS, sellOrders);
+        handleOnChange(event);
```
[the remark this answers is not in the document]
r41ph: We need to update the multiStep data as well. Please see to the latest updates where it has been modified to a more clear approach."

web-marketplace/src/pages/BuyCredits/BuyCredits.tsx
```diff
@@ -34,7 +37,30 @@ export const BuyCredits = () => {
...
+
+  useEffect(() => {
+    if (
+      !loadingSanityProject &&
+      !loadingBuySellOrders &&
+      cardSellOrders.length === 0 &&
+      ((loaded && !wallet?.address) || isBuyFlowDisabled)
+    )
+      // Else if there's no connected wallet address or buy disabled, redirect to project page
+      navigate(`/project/${projectId}`, { replace: true });
```
blushi, line 53: why do we need this here in addition to the one in `BuyCreditsForm`?

web-marketplace/src/pages/BuyCredits/BuyCredits.tsx
```diff
@@ -49,6 +75,20 @@ export const BuyCredits = () => {
...
+  useEffect(() => {
+    const localStorageData = localStorage.getItem(formModel.formId);
+    if (localStorageData) {
+      const jsonData = JSON.parse(localStorageData);
+      const paymentOption = jsonData.formValues.paymentOption;
+      const maxStep = jsonData.maxAllowedStep;
+      if (maxStep !== 0) {
+        setMaxAllowedStep(maxStep);
+      }
+      setPaymentOption(paymentOption as PaymentOptionsType);
```
blushi, line 88: why do we need this here in addition to `setPaymentOption(prev => data?.paymentOption || prev);` in BuyCreditsForm?

web-marketplace/src/pages/BuyCredits/BuyCredits.tsx
```diff
@@ -73,6 +113,13 @@ export const BuyCredits = () => {
           steps={formModel.steps}
           initialValues={{}}
           classes={{ formWrap: 'max-w-[942px]' }}
+          forceStep={
```
blushi, line 116: why do we need this in addition to the `useEffect` line 184 in BuyCreditsForm?

web-marketplace/src/pages/BuyCredits/BuyCredits.tsx
r41ph: This probably was moved from one file to another at some point and now a merge has brought it back. I will delete it.

web-marketplace/src/pages/BuyCredits/BuyCredits.tsx
r41ph: It quickly updates `paymentOption` when loading the form if there is `localStorage` data saved, avoiding this way a visual jump between sections (from step 2 to step 1) when the form was left in the retirement step.

web-marketplace/src/pages/BuyCredits/BuyCredits.tsx
r41ph: Similar to my previous comment, doing it here avoids a visual jump from retirement to payment steps in cases when the form is loaded after a user left the form in the retirement step. I believe the code that we don't need anymore is the one in `BuyCreditsForm`.

web-marketplace/src/pages/BuyCredits/BuyCredits.tsx
blushi: worth adding a comment about this then in the code

r41ph: I think so. The new `setPaymentOption` gets triggered only when the form is instantiated. The other one every time `data` changes, its purpose hasn't changed I assume.

blushi: I'm just questioning if we really need those react/jotai state variables now that the data in local storage is synced in real time but maybe we can think about that in a follow-up.
Looks like now the `useEffect` where we have `setPaymentOption(prev => data?.paymentOption || prev)` is called several times instead of just once when the user switches payment option.
We could probably update the dep array there to include only `data?.retiring, data?.paymentOption` instead of the whole `data` to mitigate that

r41ph: I have updated the deps in `BuyCredits.Form.tsx` and now it re-renders only once.

> I'm just questioning if we really need those

My first thought is that we need them to trigger a re-render if there is localStorage data because we don't have access to the multiStep data in `BuyCredits.tsx`. Happy to review whether we can optimise this if we can allocate some time for it.

S4mmyb:
> > @erikalogie @S4mmyb have a look how the currency and credits fields behave now when a user leaves the inputs with a value of zero or empty, let me know if you want to change anything.
> 
> I think this is fine? @S4mmyb wdyt

Yeah I think this is good. 

@r41ph I just tested and was noticing: 
- When I leave the buy page with the number of credits and come back the credit field is set to 1, but the currency field throws an error saying the currency must be at least 0.5 USD. It looks like it's pulling the average price per credit, which is also listed in the order price rather than the price per credit. This should be price per credit, not average price, so I'm assuming this is delt with in another PR? If not then we should be pulling the sell order price (as well as displaying it) and we shouldn't get this error if it auto-renders 1 credit if i leave the page and come back. 
- On the next page I'm noticing that there's an option for Klarna (pay over time). I might have missed this when cory was helping you design this, @erikalogie , but my first question is are we using that? If so, I'm noticing that if i select that option and go to the next page it just goes to the retirement page without asking for card info (which again if we're using it i'm assuming is in another PR). But, if i go to that page then go back the option for Klarna disappears.

blushi:
> When I leave the buy page with the number of credits and come back the credit field is set to 1, but the currency field throws an error saying the currency must be at least 0.5 USD. It looks like it's pulling the average price per credit, which is also listed in the order price rather than the price per credit. This should be price per credit, not average price, so I'm assuming this is delt with in another PR? If not then we should be pulling the sell order price (as well as displaying it) and we shouldn't get this error if it auto-renders 1 credit if i leave the page and come back.

@S4mmyb 
This is the price of the cheapest fiat sell order that is being shown as you can see on sanity:

image


And in the case where credits are only selected from one sell order, the avg price is the same as the sell order price. The issue here is that the price for one credit is lower than 0.5 USD which is the minimum accepted by Stripe. So quite an edge case, but we could have a follow-up issue to set it to 0.5 USD in this case and adjust the credits amount accordingly instead of having just 1 credits selected by default.

blushi:
> On the next page I'm noticing that there's an option for Klarna (pay over time). I might have missed this when cory was helping you design this, @erikalogie , but my first question is are we using that? If so, I'm noticing that if i select that option and go to the next page it just goes to the retirement page without asking for card info (which again if we're using it i'm assuming is in another PR). But, if i go to that page then go back the option for Klarna disappears.

I do not see this option in France. I guess this is a default option for US folks set up from Stripe, but we could disable it from Stripe dashboard since we haven't really planned for it.

erikalogie:
> > On the next page I'm noticing that there's an option for Klarna (pay over time). I might have missed this when cory was helping you design this, @erikalogie , but my first question is are we using that? If so, I'm noticing that if i select that option and go to the next page it just goes to the retirement page without asking for card info (which again if we're using it i'm assuming is in another PR). But, if i go to that page then go back the option for Klarna disappears.
> 
> I do not see this option in France. I guess this is a default option for US folks set up from Stripe, but we could disable it from Stripe dashboard since we haven't really planned for it.

Yes, let's disable it. Maybe I haven't been seeing it either because I'm in the DR, though I do use a VPN for Miami. I can see it now though.

r41ph:
> The issue here is that the price for one credit is lower than 0.5 USD which is the minimum accepted by Stripe...

@S4mmyb  @blushi I've gone ahead and fixed this issue here. Now when loading the form, when resetting the currency and credits fields on leaving them empty or with a zero value, and when changing the payment option, the currency value should not be smaller than 0.5 (when the payment option is credit card). Now we should only see the 'Must be at least 0.5 USD' when introducing manually an amount smaller than 0.5.

@blushi To do this I have done a bit of refactoring to remove some repeated code. Please review.

blushi (changes requested):

S4mmyb:
> > When I leave the buy page with the number of credits and come back the credit field is set to 1, but the currency field throws an error saying the currency must be at least 0.5 USD. It looks like it's pulling the average price per credit, which is also listed in the order price rather than the price per credit. This should be price per credit, not average price, so I'm assuming this is delt with in another PR? If not then we should be pulling the sell order price (as well as displaying it) and we shouldn't get this error if it auto-renders 1 credit if i leave the page and come back.
> 
> @S4mmyb This is the price of the cheapest fiat sell order that is being shown as you can see on sanity:
> And in the case where credits are only selected from one sell order, the avg price is the same as the sell order price. The issue here is that the price for one credit is lower than 0.5 USD which is the minimum accepted by Stripe. So quite an edge case, but we could have a follow-up issue to set it to 0.5 USD in this case and adjust the credits amount accordingly instead of having just 1 credit selected by default.

That makes sense - I didn't realize we were using Sanity to generate FIAT prices but I understand why. In the future we will probably want to migrate out of that approach to put less burden on us for managing price listings - I'll try and scope that out in a product idea spec. 

And thanks @r41ph , I tested and looks good. Though I'll wait for you to address @blushi comment to test again. 

On a separate thread, I'm curious if folks have thoughts on how we list / represent average price for these. Given that it can pull from multiple sell orders, the average price (can) update frequently as I update the credit quantity if it needs to pull from multiple sell orders. I know we show the final price at the end, but it might be worth trying to include a tool-tip explaining what the average price is and why it changes? Thoughts @erikalogie @blushi @r41ph ?

erikalogie:
> On a separate thread, I'm curious if folks have thoughts on how we list / represent average price for these. Given that it can pull from multiple sell orders, the average price (can) update frequently as I update the credit quantity if it needs to pull from multiple sell orders. I know we show the final price at the end, but it might be worth trying to include a tool-tip explaining what the average price is and why it changes? Thoughts @erikalogie @blushi @r41ph ?

In the comps, for fiat orders I use "price per credit" as I assumed since we are managing in Sanity we would only have one price, even if there are multiple sell orders. The tooltip has already been implemented for the fiat purchases. What do you think about changing the wording to "price per credit" for fiat purchases as per the comps?

S4mmyb:
> > On a separate thread, I'm curious if folks have thoughts on how we list / represent average price for these. Given that it can pull from multiple sell orders, the average price (can) update frequently as I update the credit quantity if it needs to pull from multiple sell orders. I know we show the final price at the end, but it might be worth trying to include a tool-tip explaining what the average price is and why it changes? Thoughts @erikalogie @blushi @r41ph ?
> 
> In the comps, for fiat orders I use "price per credit" as I assumed since we are managing in Sanity we would only have one price, even if there are multiple sell orders. The tooltip has already been implemented for the fiat purchases. What do you think about changing the wording to "price per credit" for fiat purchases as per the comps?

I think that would be less confusing for users, yeah. Do you agree?

blushi:
> > On a separate thread, I'm curious if folks have thoughts on how we list / represent average price for these. Given that it can pull from multiple sell orders, the average price (can) update frequently as I update the credit quantity if it needs to pull from multiple sell orders. I know we show the final price at the end, but it might be worth trying to include a tool-tip explaining what the average price is and why it changes? Thoughts @erikalogie @blushi @r41ph ?
> 
> In the comps, for fiat orders I use "price per credit" as I assumed since we are managing in Sanity we would only have one price, even if there are multiple sell orders. The tooltip has already been implemented for the fiat purchases. What do you think about changing the wording to "price per credit" for fiat purchases as per the comps?

Although I'm not entirely sure if that will be often the case in reality, it's technically possible to put different USD prices per sell orders for a same project in sanity and I think that makes sense since on chain sell orders can have different prices in crypto so strictly speaking it's still the avg price that we show, even for fiat orders.

erikalogie:
> > > On a separate thread, I'm curious if folks have thoughts on how we list / represent average price for these. Given that it can pull from multiple sell orders, the average price (can) update frequently as I update the credit quantity if it needs to pull from multiple sell orders. I know we show the final price at the end, but it might be worth trying to include a tool-tip explaining what the average price is and why it changes? Thoughts @erikalogie @blushi @r41ph ?
> > 
> > 
> > In the comps, for fiat orders I use "price per credit" as I assumed since we are managing in Sanity we would only have one price, even if there are multiple sell orders. The tooltip has already been implemented for the fiat purchases. What do you think about changing the wording to "price per credit" for fiat purchases as per the comps?
> 
> Although I'm not entirely sure if that will often be the case in reality, it's technically possible to put different USD prices per sell orders for a same project in sanity and I think that makes sense since on chain sell orders can have different prices in crypto for a same project too so strictly speaking it's still the avg price that we show, even for fiat orders.

@S4mmyb what do you want to do here? I'm pretty sure we can manage this in a way that it is always one price but wdyt?

web-marketplace/src/pages/BuyCredits/BuyCredits.utils.tsx
```diff
@@ -325,3 +300,65 @@ export const getCreditsAvailableBannerText = (
...
+  });
+
+  let newCreditsAmount = currentCreditsAmount;
+  let newCurrencyAmount = currencyAmount;
+
+  if (
+    newCurrencyAmount < MIN_USD_CURRENCY_AMOUNT &&
+    paymentOption === PAYMENT_OPTIONS.CARD
+  ) {
+    newCreditsAmount = MIN_USD_CURRENCY_AMOUNT / currencyAmount;
```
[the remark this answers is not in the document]
r41ph: Nice one. 

That made me think that there is another edge case that we need to contemplate: if the quantity of fiat credits available in a sell order (or sell orders) can not reach the minimum $0.5 needed to be able to buy with credit card, we should display a warning, or probably better just not have the option of buying with credit card available at all.

For instance:
- There is one sell order with 1 credit available at $0.2 and another sell order with 2 credits at $0.1, even buying all the credits wouldn't reach the minimum of $0.5 so it's not possible to buy them.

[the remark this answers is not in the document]
blushi: right, I think we should just have the credit card buying option disabled in this case, wdyt @erikalogie @S4mmyb?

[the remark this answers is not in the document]
blushi: in any case, this specific case should probably be tackled in a separate task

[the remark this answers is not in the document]
r41ph: I created an issue for this: https://regennetwork.atlassian.net/browse/APP-571. @S4mmyb @erikalogie feel free to update the description if needed.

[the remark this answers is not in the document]
erikalogie: Yes the credit card buying option should be disabled in this case

r41ph:
> quite an edge case, but if multiple sell orders with different prices need to be selected...

@blushi have a look at the latest update to fix this please

S4mmyb:
> > > In the comps, for fiat orders I use "price per credit" as I assumed since we are managing in Sanity we would only have one price, even if there are multiple sell orders. The tooltip has already been implemented for the fiat purchases. What do you think about changing the wording to "price per credit" for fiat purchases as per the comps?
> >   
> > Although I'm not entirely sure if that will often be the case in reality, it's technically possible to put different USD prices per sell orders for a same project in sanity and I think that makes sense since on chain sell orders can have different prices in crypto for a same project too so strictly speaking it's still the avg price that we show, even for fiat orders.
> 
> @S4mmyb what do you want to do here? I'm pretty sure we can manage this in a way that it is always one price but wdyt?

Realistically,  we won't multiple prices for fiat purchases atm as right now as we'll be managing them on-behalf of partners (i.e. project developers) who will be selling credits into a retired state. In the future when this more of a self-serve feature where any credit holder can manage selling on-chain credits with fiat we can treat this use case like we do crypto sell orders and list at multiple prices. For now though, I think we do price per credit.

web-marketplace/src/pages/BuyCredits/BuyCredits.utils.tsx
```diff
@@ -325,3 +300,110 @@ export const getCreditsAvailableBannerText = (
...
+  let newCurrencyAmount = currencyAmount;
+
+  if (
+    newCurrencyAmount < MIN_USD_CURRENCY_AMOUNT &&
+    paymentOption === PAYMENT_OPTIONS.CARD
+  ) {
+    newCreditsAmount = calculateCredits(
+      orderedSellOrders as CardSellOrder[],
+      MIN_USD_CURRENCY_AMOUNT,
+    );
```
blushi, line 348: sell orders (to be set below in `setValue(SELL_ORDERS, ...)`) should be recomputed too
also can't we just use the existing function `getCreditsAmount` with `value` param equal to `MIN_USD_CURRENCY_AMOUNT`? this also returns the new sell orders

r41ph: Thanks - not sure what I was thinking!

r41ph: @blushi see refactored code using `getCreditsAmount` please

PR #2576, opened 2025-01-15
fix: APP-425 remove walletLoaded prop from LedgerProvider

r41ph:
## Description

https://regennetwork.atlassian.net/browse/APP-425

### Author Checklist

I have...

### How to test

1. Go to https://deploy-preview-2576--regen-marketplace.netlify.app/.
2. Check that the Featured projects and Credit classes sections are loading as expected.
3. Log in using Keplr extension and then lock the wallet as explained here: https://regennetwork.atlassian.net/browse/APP-425?focusedCommentId=10856
4. Refresh and verify that the Featured projects and Credit classes load as expected.

### Reviewers Checklist

I have...

- [ ] reviewed tests


web-marketplace/src/ledger.tsx
```diff
@@ -83,26 +83,21 @@ const getApi = async (
...
+> = ({ wallet, children }) => {
   const [api, setApi] = useState<RegenApi | undefined>(undefined);
   const [loading, setLoading] = useState<boolean>(false);
   const [error, setError] = useState<unknown>(undefined);
 
   useEffect(() => {
-    if (walletLoaded && !api)
-      getApi(setApi, setLoading, setError, wallet?.offlineSigner);
-  }, [api, setApi, setLoading, setError, wallet?.offlineSigner, walletLoaded]);
+    if (!api) getApi(setApi, setLoading, setError, wallet?.offlineSigner);
```
blushi, line 99: this will cause another bug: since connecting to keplr wallet is done async, `wallet?.offlineSigner` will be undefined initially, which means we will instantiate `api` without the signer (even though there could be one)
then once `wallet?.offlineSigner` gets defined, we won't reset the `api` because it's already set (`if (!api)`)
meaning that you won't be able to sign transactions although your wallet is connected

I believe we could check in addition the following: `!!wallet?.offlineSigner && !api.msgClient` (`msgClient` is undefined if there's no signer: https://github.com/regen-network/regen-js/blob/v0.7.1/packages/api/src/api.ts#L66)

r41ph: @blushi see whether the latest updates sort this issue please.

> I believe we could check in addition the following..

Should we create a new issue to update this?

r41ph:

web-marketplace/src/ledger.tsx
blushi: just using the following instead does the same thing as the code updates you added:
```suggestion
    if (!api || (!!wallet?.offlineSigner && !api.msgClient)) getApi(setApi, setLoading, setError, wallet?.offlineSigner);
```
but it looks simpler since we don't need a new `useEffect` nor a new state variable

web-marketplace/src/ledger.tsx
```diff
@@ -96,13 +96,23 @@ export const LedgerProvider: React.FC<
...
 
+  // Get API even if wallet is not loaded to unblock data fetching.
   useEffect(() => {
-    if (walletLoaded && !api)
+    if (!api) getApi(setApi, setLoading, setError, wallet?.offlineSigner);
+  }, [api, setApi, setLoading, setError, wallet?.offlineSigner]);
+
+  // Update API when the wallet is loaded.
+  useEffect(() => {
+    if (api && !apiUpdatedWithOfflineSigner && walletLoaded) {
```
blushi, line 111: `walletLoaded` doesn't necessarily mean a wallet has been connected
if there's no wallet to connect to, then `walletLoaded` is set from false to true in https://github.com/regen-network/regen-web/blob/dev/web-marketplace/src/lib/wallet/hooks/useAutoConnect.tsx#L44
which means we don't need to reinstantiate the api in this case

blushi (approved):
tACK
Reminder to not merge from github
Side note that in the `dev` sync PR, we might need to handle this (which is not yet on `main` but might cause similar issues in the next `dev`->`main` release): https://github.com/regen-network/regen-web/blob/fca9f0521d2ef03cc78a36603bb2f06d09d93159/web-marketplace/src/ledger.tsx#L91

erikalogie:
LGTM

S4mmyb:
Just tested @r41ph and looks good to me

PR #2577, opened 2025-01-16
feat: APP-399 "how to purchase credits" popup when user is not logged in with keplr and no fiat credits available

blushi:
## Description

Closes: https://regennetwork.atlassian.net/browse/APP-399

Buying options filter now appear in the URL query params so its state is entirely handled from react-router search params so it's not duplicated with jotai state

---

### Author Checklist

I have...

### How to test

1. Go to https://deploy-preview-2577--regen-marketplace.netlify.app/project/C02-002 (without being logged in or logged in with a web2 account)
2. Test "connect wallet to buy" button, after successful wallet connection, this should redirect to the project buy page
3. Test "or, find credit card projects" button: this should redirect to the /projects page with credit card filter enabled

### Reviewers Checklist

I have...

- [ ] reviewed tests


blushi:

I've made it so that the buying options filters are URL query parameters, this way it's easy to get the /projects link with those filters enabled.
I haven't done it for the other existing filters (credit class and community projects) since this was a bit out of scope but let me know if you want to do the same thing for them and I'll create a follow-up issue. One question I'm wondering about is: how do we want to handle credit class filter since all credit classes are checked by default? I think we don't want to have all of the selected credit classes defined in the url in this case.

erikalogie:
> @erikalogie @S4mmyb see testing instructions
> 
> I've made it so that the buying options filters are URL query parameters, this way it's easy to get the /projects link with those filters enabled. I haven't done it for the other existing filters (credit class and community projects) since this was a bit out of scope but let me know if you want to do the same thing for them and I'll create a follow-up issue. One question I'm wondering about is: how do we want to handle credit class filter since all credit classes are checked by default, except our "skipped credit class"? I think we don't want to have all of the selected credit classes defined in the url in this case.

I think it would be good to have a follow up issue. No I agree we shouldn't have all the selected credit classes defined in the url. Do you have suggestions for simplifying?

This is working as expected. This will work will be further enhanced on mobile when we implement this task so that you can see the filter count on the button: https://regennetwork.atlassian.net/jira/software/c/projects/ENG/boards/51/backlog?selectedIssue=APP-545

S4mmyb:
This LGTM, thanks for splitting out that other issue.

blushi:
@r41ph could you have a look at this please?

r41ph (changes requested):

r41ph:
The issue may be here: https://github.com/regen-network/regen-web/pull/2577/files#diff-c3146a40b7edd942cdadcb1a51b5090c4c0bbe3b160359329299f14263bc1b31R169

We may need to do add the filter query to the redirection:

```javascript
useEffect(() => {
// As soon as a filter changes, we navigate back to page 1 if on all projects page
 if (routePage && Number(routePage) > 1) {
    const buildBuyingOptionsQueryString = (
        filters: Record<string, boolean>,
      ): string => {
        const params = Object.keys(filters)
          .map(key => `buying_options_filters=${key}`)
          .join('&');

        return params ? `?${params}` : '';
      };
      const queryString = buildBuyingOptionsQueryString(buyingOptionsFilters);
      navigate(`/projects/1${queryString}`);
...
```

blushi:
> The issue may be here: https://github.com/regen-network/regen-web/pull/2577/files#diff-c3146a40b7edd942cdadcb1a51b5090c4c0bbe3b160359329299f14263bc1b31R169
> 
> We may need to do add the filter query to the redirection:
> 
> ```js
> useEffect(() => {
> // As soon as a filter changes, we navigate back to page 1 if on all projects page
>  if (routePage && Number(routePage) > 1) {
>     const buildBuyingOptionsQueryString = (
>         filters: Record<string, boolean>,
>       ): string => {
>         const params = Object.keys(filters)
>           .map(key => `buying_options_filters=${key}`)
>           .join('&');
> 
>         return params ? `?${params}` : '';
>       };
>       const queryString = buildBuyingOptionsQueryString(buyingOptionsFilters);
>       navigate(`/projects/1${queryString}`);
> ...
> ```

Yep good catch

r41ph (approved):

PR #2641, opened 2025-03-26
fix: APP-633 credit batch creation amounts shift

r41ph:
## Description

https://regennetwork.atlassian.net/browse/APP-633

### Author Checklist

I have...

### How to test

1. Verify that all numeric input fields in the marketplace do not change their value when focused and the user scrolls with the mouse wheel.

### Reviewers Checklist

I have...

- [ ] reviewed tests

### DIFF
--- a/web-components/src/components/inputs/hooks/usePreventWheelScroll.tsx
+++ b/web-components/src/components/inputs/hooks/usePreventWheelScroll.tsx
@@ -1,0 +1,41 @@
+import { RefObject, useEffect } from 'react';
+
+/**
+ * Hook that prevents the default scroll behavior on mouse wheel events
+ * for the provided element reference when it's focused
+ *
+ * @param elementRef - A React ref object pointing to a DOM element
+ */
+export default function usePreventWheelScroll(
+  elementRef: RefObject<HTMLElement>,
+  shouldPreventScroll = true,
+): void {
+  useEffect(() => {
+    const currentRef = elementRef.current;
+    if (!currentRef || !shouldPreventScroll) return;
+
+    const preventScrollOnWheel = (e: WheelEvent) => {
+      e.preventDefault();
+    };
+
+    const handleFocus = () => {
+      currentRef.addEventListener('wheel', preventScrollOnWheel, {
+        passive: false,
+      });
+    };
+
+    const handleBlur = () => {
+      currentRef.removeEventListener('wheel', preventScrollOnWheel);
+    };
+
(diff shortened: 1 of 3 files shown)


r41ph:

erikalogie:
LGTM!

r41ph:
@beccaRND I've noticed in your video that you are testing this in dev.app.regen.network. There is a fix deployed for this here: https://deploy-preview-2641--regen-marketplace.netlify.app/. Can you try over there please?

Also, this issue happens when using the mouse wheel to scroll when the 'amount tradable' input is focused as moving the wheel changes the amount. Can you confirm this is what is happening to you too?

web-components/src/components/inputs/hooks/usePreventWheelScroll.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Hook that prevents the default scroll behavior on mouse wheel events
+ * for the provided element reference when it's focused
+ *
+ * @param elementRef - A React ref object pointing to a DOM element
+ */
+export default function usePreventWheelScroll(
+  elementRef: RefObject<HTMLElement>,
+  shouldPreventScroll = true,
```
blushi, line 11: why do we need shouldPreventScroll here? doesn't seem to be used

blushi (approved):
utACK, pending one question

web-components/src/components/inputs/hooks/usePreventWheelScroll.tsx
r41ph: `shouldPreventScroll` helps avoid setting the event listeners on every input field where we use `usePreventWheelScroll`. We only need to prevent the scrolling for inputs `type=number`, so in both `<TextField/>` components we pass `shouldPreventScroll` based on the input type.

See line 15 https://github.com/regen-network/regen-web/pull/2641/files#diff-ea7d3c3e2dfcaf7145c8962fd69a81d2f46e0965527c4bdcce9795e731432b9dR15

blushi: oh I see, I was just looking for `shouldPreventScroll` string but it's a positional parameter so it's named something else
I think it would be useful to add some doc about this param too in `usePreventWheelScroll` docs

r41ph: ah yes, I have updated the hook docs

PR #2650, opened 2025-04-23
fix: APP-642 browser back button in MultiStep 

r41ph:
## Description

Closes: #XXXX or regen-network/rnd-dev-team#XXXX

---

### Author Checklist

I have...

### How to test

1.

### Reviewers Checklist

I have...

- [ ] reviewed tests


web-components/src/components/header/components/HeaderLogoLink.tsx
```diff
@@ -1,5 +1,4 @@
 import React from 'react';
-import NextLink from 'next/link';
```
r41ph, line 2: Not related, just some clean up

web-components/src/components/header/index.tsx
```diff
@@ -9,7 +9,7 @@ import {
   HeaderMenuItem,
   Item,
 } from './components/HeaderMenuItem/HeaderMenuItem';
-import { NavLink, NavLinkProps } from './components/NavLink';
+import { NavLink } from './components/NavLink';
```
r41ph, line 12: Not related, just some clean up

web-marketplace/src/clients/regen/Regen.Routes.tsx
```diff
@@ -165,6 +165,7 @@ export const getRegenRoutes = ({
         />
         <Route path="verify-email" element={<VerifyEmail />} />
         <Route path="add" element={<Additionality />} />
+        {/* TODO: should this route be deleted? */}
```
r41ph, line 168: I wonder if we can delete this redirection and the one below or are they still needed?

r41ph:

r41ph:
@erikalogie I have updated this with a fix for when projects have a slug, have a look please.

erikalogie:
The same thing is still happening for me:

https://github.com/user-attachments/assets/7a374845-66bf-4383-962e-e82cf35f6331

r41ph:
That project doesn't have a slug, what is exactly the issue here please? it looks like you click back in the browser and it doesn't seem to always work?

blushi:
> @blushi, I looked into the refactoring we talked about the other day and changing the Buy Credits flow to use routes is a lot of work and testing just to only improve the navigation UX (specially if things are moving for the migration to Next.js). We talked about all this in the last standup, when you were away, and agreed to discuss it in the next sprint planning meeting.

Like we talked about in our 1-1, I agree it’s best to hold off on any big route changes until after the Next.js migration.

> So I haven't done much with this PR since, but I have polished it a bit in case we can see an improvement in the UX that is worth deploying, even if not perfect it can be a better UX, what do you all think? @erikalogie @S4mmyb

When I go to a project buy page from another page (eg /projects) and click the browser back button, then it stays on the buy page but without the sub-route, ie /project/C50-001/buy instead of /project/C50-001/buy/choose-credits, which is not really an improvement compared to before, unless we can fix it easily? Also if I go through the steps and then go back using the browser back button, somehow I'm blocked on the Customer info form, the browser back button is disabled. Given the above, not sure if it's worth pursuing the changes here.

r41ph:
OK - I thnk we should leave this for now because it's creating more issues than it solves, and wait for the migration to Next.js. Then we can decide if we want to update the MultiStep component to work with routes or, perhaps, get rid of it completely and use just normal routes to handle multi step forms and reuse the "step counter" that shows which is the current active step.

blushi:
@r41ph should we just close this then?

PR #2662, opened 2025-05-21
feat: next migration setup

blushi:
## Description

APP-688
APP-689

We use a [catch all segment](https://nextjs.org/docs/app/building-your-application/routing/dynamic-routes#optional-catch-all-segments) in the app router to serve the existing app with react-router.
The diff is pretty large given I had to move lots of file:
- /pages to /legacy-pages to not conflict with the page router folder 
-  packages upgrades, mainly lingui, material-ui and react-crop-image

So I'd recommend to test locally rather than reading it all because the changes are mostly repetitive (best to delete all /node_modules folders and run a fresh `bun i`)

There are still some fast refresh issues on the react-router routes (some part of the styles, font in particular, not applied on automatic refresh), which doesn't appear on pure app router route so I didn't dig too much into it for now.

We're not using turbopack yet, but it could be a follow-up issue if we feel like the DX is not the best. For now, part of the reason might just be that we have to compile all the routes initially within this catch all segment.

---

### Author Checklist

I have...

### How to test

1.

### Reviewers Checklist

I have...

- [ ] reviewed tests


web-components/package.json
```diff
@@ -30,12 +30,11 @@
     "@mapbox/togeojson": "^0.16.0",
     "@math.gl/core": "^4.0.0",
     "@math.gl/web-mercator": "^4.0.0",
-    "@mui/icons-material": "^5.1.0",
     "@mui/lab": "^5.0.0-alpha.54",
-    "@mui/material": "^5.10.10",
-    "@mui/styled-engine": "^5.10.8",
-    "@mui/styles": "^5.10.10",
-    "@mui/system": "^5.10.10",
+    "@mui/material": "^5.16.12",
```
blushi, line 34: this was to fix this https://github.com/mui/material-ui/issues/41388#issuecomment-2506157151

r41ph (approved):
I think this is a solid start for the Next.js migration. Using the catch all route for the legacy app allows us for a incremental transition which is great, and although the DX is not great at the moment (fast refresh issues you already mentioned), the production app seems to run well. A couple of things:

- We have duplicated styles imports for `App.css` and `tailwind.css` - I think we should remove the ones in `Regen.Providers.tsx` and leave the ones in `layout.tsx`.
- Do we need to keep the `vite.config.mts` file and all the Vite deps in the package.json?

blushi:
> * Do we need to keep the `vite.config.mts` file and all the Vite deps in the package.json?

see https://github.com/regen-network/regen-web/pull/2662#discussion_r2101977483, it's still used for the test and storybook so keeping them for now

blushi:
> * We have duplicated styles imports for `App.css` and `tailwind.css` - I think we should remove the ones in `Regen.Providers.tsx` and leave the ones in `layout.tsx`.

I think I'm actually gonna keep App.css in layout.tsx but keep `tailwind.css` in `Regen.Providers.tsx` because for terrasos, we have a different css.
I've created a separate task to handle the conditional terrasos providers/css https://regennetwork.atlassian.net/browse/APP-722

PR #2696, opened 2025-07-24
chore: APP-736 portfolio empty state

blushi:
## Description

https://regennetwork.atlassian.net/browse/APP-736

---

### Author Checklist

I have...

### How to test

1. Log in with an account that has not ecocredits and check /dashboard/portfolio and your public profile portfolio
2. Log in with an account with some ecocredits to check the portfolio hasn't changed

### Reviewers Checklist

I have...

- [ ] reviewed tests


web-marketplace/src/components/organisms/EcocreditsTable.tsx
```diff
@@ -117,49 +117,54 @@ export const EcocreditsTable: React.FC<
...
-      rows={credits.map((row, i) => {
-        return [
-          <WithLoader isLoading={row.projectName === ''} variant="skeleton">
-            <Link
-              href={`/project/${row?.projectId}`}
-              sx={tableStyles.ellipsisColumn}
+      rows={
+        credits?.map((row, i) => {
+          return [
+            <WithLoader isLoading={row.projectName === ''} variant="skeleton">
```
r41ph, line 123: I believe we need to add `key` props to this array elements.

r41ph (changes requested):

erikalogie:
> > https://www.loom.com/share/8d6578937a4942799ec3037fca0ef05b
> 
> ok so we should use 10px border radius everywhere what about the header height (where it's written "Portfolio" or "Sell")? Currently they are different, which should be fixed too I guess, if so, which height to keep

I like the height of "Portfolio" better, and I think it is more like the comps

S4mmyb:
@blushi, I'm noticing an error for an account that has credits. It looks like the UI is showing the unfiltered gRPC response from `regen.ecocredit.v1.Query/Balances`, which includes all batches that the account has ever interacted with — not just those where a balance is currently held. For example, the address `regen1v2ncquer9r2ytlkxh2djmmsq3e8we6rjc9snfn` returns 12 entries, but only one has non-zero credit amounts.

I'm realizing that that's the issue I was running into on prod too. That same address should display a balance from 3 batches, but shows 14 and splits displaying those on both pages. 

https://github.com/user-attachments/assets/87166d5b-ce32-4291-81d1-650c0afe9e47

blushi:
> @blushi, I'm noticing an error for an account that has credits. It looks like the UI is showing the unfiltered gRPC response from `regen.ecocredit.v1.Query/Balances`, which includes all batches that the account has ever interacted with — not just those where a balance is currently held. For example, the address `regen1v2ncquer9r2ytlkxh2djmmsq3e8we6rjc9snfn` returns 12 entries, but only one has non-zero credit amounts.
> 
> I'm realizing that that's the issue I was running into on prod too. That same address should display a balance from 3 batches, but shows 14 and splits displaying those on both pages.
> 
>  Recording.2025-07-30.220131.mp4

@S4mmyb this should be covered by https://regennetwork.atlassian.net/jira/software/c/projects/ENG/boards/51?selectedIssue=APP-744 see my comments there

blushi:
> > Looking at the sell dashboard with some sell orders, the cards border radius is actually 5px there:
> > image
> > so should I change it to 10px too or not? On figma, it's all set to 5px, except for the project cards.
> 
> Let's do everything 10px then and I'll fix it in Figma.

you can have another look @erikalogie

erikalogie:
Monosnap Regen Marketplace : Buy, trade and retire carbon credits | Regen Marketplace : Buy, trade and retire carbon credits 2025-07-31 13-59-54

Sorry one small comment

web-marketplace/src/components/organisms/EcocreditsTable.tsx
blushi: oh I see, I don't think we need keys in this case, we are just mapping `credits` onto an array of elements (which are table cells content essentially) which then will be embedded and keys are used in this case

blushi: I haven't actually modify the code logic, just handling potential undefined `credits`

blushi:
> Monosnap Regen Marketplace : Buy, trade and retire carbon credits | Regen Marketplace : Buy, trade and retire carbon credits 2025-07-31 13-59-54
> Sorry one small comment

@erikalogie you can have another look

erikalogie:
@blushi looks great

PR #2720, opened 2025-08-22
feat: APP-772 follow up comments for 674

alexander-astrand:
## Description

https://regennetwork.atlassian.net/browse/APP-772

Comments here:
https://github.com/regen-network/regen-web/pull/2718/files

---

### Author Checklist

I have...

### How to test

1.

### Reviewers Checklist

I have...

- [ ] reviewed tests

### DIFF
--- a/web-marketplace/src/components/atoms/ToggleSwitch.tsx
+++ b/web-marketplace/src/components/atoms/ToggleSwitch.tsx
@@ -1,0 +1,112 @@
+import React, { useId } from 'react';
+
+import { cn } from 'web-components/src/utils/styles/cn';
+
+/* eslint-disable lingui/no-unlocalized-strings */
+const SIZE_CONFIG = {
+  sm: {
+    height: 'h-22',
+    knob: 'w-16 h-16',
+    font: 'text-xs',
+    paddingX: { on: 'pl-3 pr-2', off: 'pl-2 pr-3' },
+    paddingY: 'py-2',
+    gap: 'gap-4',
+    knobMargin: { on: 'ml-3', off: 'mr-3' },
+  },
+  md: {
+    height: 'h-28',
+    knob: 'w-20 h-20',
+    font: 'text-[13.3333px]',
+    paddingX: { on: 'pl-5 pr-3', off: 'pl-3 pr-5' },
+    paddingY: 'py-3',
+    gap: 'gap-5',
+    knobMargin: { on: 'ml-4', off: 'mr-4' },
+  },
+} as const;
+
+export interface ToggleSwitchProps {
+  checked: boolean;
+  onChange: (checked: boolean) => void;
+  disabled?: boolean;
+  onLabel?: React.ReactNode;
+  offLabel?: React.ReactNode;
+  ariaLabel?: string;
+  className?: string;
+  size?: 'sm' | 'md';
+}
+
+export const ToggleSwitch: React.FC<ToggleSwitchProps> = ({
+  checked,
+  onChange,
+  disabled = false,
+  onLabel,
+  offLabel,
+  ariaLabel,
+  className,
+  size = 'md',
+}) => {
+  const id = useId();
+  const cfg = SIZE_CONFIG[size];
+  const paddingX = checked ? cfg.paddingX.on : cfg.paddingX.off;
+  const {
+    height,
+    knob: knobSize,
+    font: fontSize,
+    paddingY,
+    gap,
+    knobMargin,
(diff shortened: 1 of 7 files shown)


web-marketplace/src/components/atoms/ToggleSwitch.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+export interface ToggleSwitchProps {
+  checked: boolean;
+  onChange: (checked: boolean) => void;
+  disabled?: boolean;
+  onLabel?: React.ReactNode;
+  offLabel?: React.ReactNode;
+  ariaLabel?: string;
+  className?: string;
+  size?: 'sm' | 'md';
```
r41ph, line 35: We could use the keys defined in SIZE_CONFIG `size?: keyof typeof SIZE_CONFIG;` so if we add another size we won't need to update the `ToggleSwitchProps` interface.

web-marketplace/src/components/atoms/ToggleSwitch.tsx
```diff
@@ -0,0 +1,112 @@
...
+        type="checkbox"
+        role="switch"
+        aria-checked={checked}
+        aria-label={ariaLabel}
+        className="sr-only"
+        checked={checked}
+        disabled={disabled}
+        onChange={e => !disabled && onChange(e.target.checked)}
+      />
+      {checked ? (
```
r41ph, line 85: I believe we could improve this by avoid unmounting/remounting when the `checked` state changes. Maybe we could use a single DOM structure here for the knob and label? that way we could add an animation to the toggle if we want to

r41ph (commented):
This looks good - I have added a couple of comments to the `ToggleSwitch.tsx` after having another look.

web-marketplace/src/components/atoms/ToggleSwitch.tsx
```diff
@@ -0,0 +1,112 @@
...
+  onChange: (checked: boolean) => void;
+  disabled?: boolean;
+  onLabel?: React.ReactNode;
+  offLabel?: React.ReactNode;
+  ariaLabel?: string;
+  className?: string;
+  size?: 'sm' | 'md';
+}
+
+export const ToggleSwitch: React.FC<ToggleSwitchProps> = ({
```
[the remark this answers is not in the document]
alexander-astrand: Agreed, fixed it now!

web-marketplace/src/components/atoms/ToggleSwitch.tsx
```diff
@@ -0,0 +1,112 @@
+import React, { useId } from 'react';
+
+import { cn } from 'web-components/src/utils/styles/cn';
+
+/* eslint-disable lingui/no-unlocalized-strings */
+const SIZE_CONFIG = {
+  sm: {
```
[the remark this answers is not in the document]
alexander-astrand: Started out building for different sizes but looks like we won't need it now, I removed it now and we can add that feature later if we were to need it!

web-marketplace/src/components/atoms/ToggleSwitch.tsx
erikalogie: Would love it to behave like in the mantime example if possible, not sure how difficult that is. 

Good call on making the on/off states the same width, something I missed in the designs!

r41ph: Looks much better. Do we need to use `'framer-motion'` for such a small animation?

r41ph (approved):
Approving - I'm just not sure about using 'framer-motion' in this scenario ([see](https://github.com/regen-network/regen-web/pull/2720#discussion_r2330255518)). 

The library is fairly big and I wonder if we can remove it altogether and see if we can migrate the current use cases to css/tailwind animations? do you think this is worth checking at some point @blushi?

blushi:
> Approving - I'm just not sure about using 'framer-motion' in this scenario ([see](https://github.com/regen-network/regen-web/pull/2720#discussion_r2330255518)).
> 
> The library is fairly big and I wonder if we can remove it altogether and see if we can migrate the current use cases to css/tailwind animations? do you think this is worth checking at some point @blushi?

We use it in different components already so it's worth a separate task I think

web-marketplace/src/components/atoms/ToggleSwitch.tsx
alexander-astrand: @r41ph I have refactored without framer-motion now!

